PR #223, opened 2014-11-12
Expose deserialization order for groups in Java

yaitskov:
During development I've written a deserializer messages to json for generated classes from XML.
And I've found that Java classes requires specific order for calling methods deserializing nested groups.

It looks weird for me. I decided to annotate generated classes with order of groups in a message.

With this feature you can get pretty printed JSON representation a messages automatically.
I think this is very usefull for development and debugging easily get representation of a problem messages.

XML spec gives you only a structure. This does human representation of real data examples.

Gson jar is not packed with sbe.jar. It used only for compilation and tests.

### DIFF
--- a/main/java/uk/co/real_logic/sbe/codec/java/MethodSelector.java
+++ b/main/java/uk/co/real_logic/sbe/codec/java/MethodSelector.java
@@ -20,9 +20,10 @@ public class MethodSelector
     private final Set<String> ignoredMethods;
     private final Map<Class, Set<String>> sortedMethods = new HashMap<>();
 
-    public static Set<String> objectMethods()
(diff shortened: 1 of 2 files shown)


juddgaddie:
@yaitskov what's the advantage of this over [uk.co.real_logic.sbe.json.JsonPrinter](https://github.com/real-logic/simple-binary-encoding/blob/2e13c06756d13e303d8463e7ce86dafd6718df32/sbe-tool/src/main/java/uk/co/real_logic/sbe/json/JsonPrinter.java) ?

mjpt777:
The `JsonPrinter` is only a partial implementation. I'd like to remove it or finish it off when I get some time. @RichardWarburton put it in for debugging at one stage.

PR #234, opened 2015-01-07
Correcting cpp targets and xsd validation

DainDwarf:
Hello,

While trying to use this library, I had trouble compiling for cpp98. By tracking down the issue, I modified some .xml example files to take into account the ```main/resources/fpl/SimpleBinary1-0.xsd``` validation given in that library, and made some ```-Doptions``` for the call on ```sbe.jar``` optional for the ```macrdef sbetool``` in ```build.xml``` (otherwise, ant would call ```sbe.jar``` with ```-Doptions=""```, which then fails to create the correct directory when using ```-Dsbe.target.namespace=""```)

Anyway, cpp examples and build seems more correct now, and it compiles for me. I hope you will find those changes usefull.


tmontgomery:
This has been pretty much changed and cleaned up. Thanks!

PR #236, opened 2015-01-22
Enable different buffer implementations to be used

RichardWarburton:
Replace the `DirectBuffer` with the one from `Agrona` and parameterises the templates accordingly in order to improve interoperability with other tools like Aeron. Current state of play is that the hand-written code depends on interfaces, but generated code can depend on any class that implements the `MutableDirectBuffer` interface.

Also pulls in @tmontgomery 's work on converting the build to gradle.

I don't want to merge these changes yet, because the C++ build is broken and I'm not sure what the appropriate way of testing the c# code is. Just pull-requesting it for comments.


RichardWarburton:
I think the native build is ok now.

tmontgomery:
BTW, I have not tried a Windows native build yet. The move to ExternalProject for gtest and other things could make that break. Let me know what you find.

PR #249, opened 2015-03-29
Single file generator for Visual Studio

ChrisShain:
Added a project to the visual studio solution that creates a single file generator (https://msdn.microsoft.com/en-us/library/bb166817.aspx) to convert .xml SBE schemas to generated code in the IDE.  The generator is published to the VS Gallery at https://visualstudiogallery.msdn.microsoft.com/f2663052-198d-42a2-97e6-25c74188bd19?SRC=Home


mjpt777:
Thanks Chris. Some of the guys are looking at this.

LeeCampbell:
Hi Chris,
I love the idea but I have a couple of things I think need to be done before this could be merged in.

 * The VisX needs to work on VS2013. The rest of the solution targets VS2013 (as the current build of Visual Studio) but I don't believe the [artefact](https://visualstudiogallery.msdn.microsoft.com/f2663052-198d-42a2-97e6-25c74188bd19?SRC=Home) works in VS2013/v12.0
 * Prior to this PR the sln could be opened in a fairly plain install of VS. With this it now errors if the VS SDK is not installed. If this is now a requirement, it should be documented somewhere.
 * The `.gitignore` file should fall into line with the current layout. You have VS exclusions in the `#mac` section. 
 * Even better than fixing the `.gitignore` would be to direct the output of the build to the `.\target\csharp` folder. Maybe `.\target\csharp\vsix`?
 * I don't think that the `sbe.jar` file should be checked in as source code. It is the output of building the main part of the repo. 
 * Some namespace confusion going on here with `Adaptive` + `ChristopherShain`

So I think that they are some blockers to getting this PR in as it stands.

The next question is, should this tool belong in the main repo or in a standalone repo. I think currently I am leaning towards it being in another repo. This has the benefit of keeping the core focus of this repo as tight as possible. It also seems to be a nice solution for the some of the issues above. The new repo could document that the VS SDK is a prerequisite. It could also simply use Nuget to pull in the .NET parts and the `sbe.jar` removing the need for it to be a binary in the source control system. Being standalone also means we can add issues like "Support VS2015" and patch them without Martin/Richard/Todd having to accept the pull-request.

If you are interested, I would be keen to contribute to this standalone repo? We could link to it from this main repo so people know where to get it.

ChrisShain:
Agreed on all points:

- The VSIX has now been updated and tested in VS2013 (that was a mistake on
my part in the VSIX manifest file- there was nothing VS2013 unfriendly
about it).  The new copy is now on VS Gallery.

- The VS SDK is an unfortunate requirement of VSIX projects.  Happy to
document it or move codegen to a different solution or even a different
repo, as you prefer.

- I've made the suggested targeting change, so the .gitignore has been
reverted to what it was before I touched it.

- For the SBE jar dependency, I've done something that I hope isn't "too
clever".  As the VSIX project already has a nuget dependency on the jar,
I've removed the checked in jar and re-added it as a link to the jar in the
packages folder under the project.  The thinking here is twofold- one, I
dont want .NET developers to have to build the java project first.
Secondly, this would make it easier to break out the VSIX project into it's
own solution later on, or to change the directory structure relative to the
target output, or to move to another repo altogether.

- Fixed all namespaces to be Adaptive.SimpleBinaryEncoding.CodeGen

Because I've committed the sin here of checking in a binary (sbe.jar) I am
thinking that I should delete my fork repository, then re-fork and re-add
my project- that is, unless you want to make a new repo and then I can add
the project there.  Let me know what you think.

Chris

mjpt777:
C# currently not supported for RC3.

Issue #221, opened 2014-11-04
Add String convenience methods for var data


mjpt777:
Add convenience methods to get and put a String for var data fields containing characters.

David-Wales:
Am I right in my understanding of this ticket;

That char datatypes specified in the Codec's XML file should generate an equivalent overloaded put method that takes a String as an argument instead of a byte array and an equivalent get method that returns a string instead of using a for loop to return by character. 

for example 

```
<type name="VehicleCode" primitiveType="char" length="6"/>
```


Pehaps generating some method as such:-
``` 
//Maybe not need to srcOffset argument in this case
public Car putVehicleCode(final String src, final int srcOffset)
```

and some kind of method on the codec that allows the following to be circumvented
```
        for (int i = 0, size = Car.vehicleCodeLength(); i < size; i++)
        {
            sb.append((char)car.vehicleCode(i));
        }
```


Thank you

tmontgomery:
You are correct. This is a usability enhancement to make it easier to use. It is more akin to grabbing the bytes into a temporary byte buffer, though. Then constructing the string from that.

mjpt777:
I added the String convenience methods for var data. Maybe it should also be added for char arrays.

PR #255, opened 2015-04-21
Unit test for BitUtil Class and Issue #221 Addition of Convenience methods for get/pur char datatypes

David-Wales:
Adding a unit test for a small utility in the Java version of the project And issue #221.  First commit so any feedback would be greatly welcomed!

### DIFF
--- a/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -278,12 +278,11 @@ public class JavaGenerator implements CodeGenerator
             bufferImplementation
         ));
 
-        sb.append(String.format(
-            indent + "    public static int sbeHeaderSize()\n" +
-            indent + "    {\n" +
-            indent + "        return HEADER_SIZE;\n" +
-            indent + "    }\n\n"
-        ));
+        sb.append(indent).append("    public static int sbeHeaderSize()\n")
+          .append(indent).append("    {\n")
+          .append(indent).append("        return HEADER_SIZE;\n")
+          .append(indent).append("    }\n\n");
+
         sb.append(String.format(
             indent + "    public static int sbeBlockLength()\n" +
             indent + "    {\n" +


examples/java/uk/co/real_logic/sbe/examples/ExampleUsingGeneratedStub.java
```diff
@@ -119,7 +119,7 @@ public static int encode(final Car car, final UnsafeBuffer directBuffer, final i
            .modelYear(2013)
            .available(BooleanType.TRUE)
            .code(Model.A)
-           .putVehicleCode(VEHICLE_CODE, srcOffset);
```
mjpt777: The example was to show the difference between using Strings and raw byte arrays. The Strings are more usable but introduce garbage. Many in HFT want their applications to be garbage free to avoid the subsequent pauses.

test/java/uk/co/real_logic/sbe/codec/java/BitUtilTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+
+import org.junit.Before;
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+public class BitUtilTest
```
mjpt777: Why reset this way rather than call the wrap methods?

examples/java/uk/co/real_logic/sbe/examples/ExampleUsingGeneratedStub.java
David-Wales: That makes sense. I can remove this from the example. Does the generator code for these convenience method look ok?

Thank you!

test/java/uk/co/real_logic/sbe/codec/java/BitUtilTest.java
David-Wales: I'm not sure.  I wrote a small test for it as it was the smallest component I could initially find. But on closer inspection I'm not sure this class (or this method) is even used anywhere.

I could remove the class and test if that would be preferable?

Cheers!

mjpt777: Removing this makes sense.

David-Wales:
So I've removed the unused BitUtil class and tests I added for it. I've also reverted the main examples back to using byte by byte access to char data types. I have left the extra code on the Java Generator  that generates these convenience methods though, as per the issue.

RichardWarburton:
I think theres a bitutil in agrona which should be used. So maybe its worth
looking at that?

mjpt777:
I think there is a major flaw with treating the char arrays as strings in SBE. The spec says a `characterEncoding` attribute can be set, it defaults to "UTF-8". Since the array is fixed size and UTF-8 is a variable byte char encoding then this does not match. It feels like char arrays should be restricted to ASCII to ensure a one-for-one byte to char mapping.

Maybe @donmendelson has a view on this?

donmendelson:
Some issues with character arrays and strings were clarified in the RC3 spec (I expect it soon to be named final draft of SBE version 1.0.)

Section 2.7.2.1	"Encoding specifications for fixed-length character array"
````
US-ASCII is the default encoding of character arrays to conform to usual FIX values. The characterEncoding attribute may be specified to override encoding. 
````

PR #256, opened 2015-04-21
Fix null pointer exception and CSharp compiler error

warrenfalk:
An empty messageHeader composite throws a NullPointerException because even though getAttributeValueOrNull() says that elementNode can be null, no null check is done on the element or the attribute map.

Also the CSharp generated code will not compile if you have a uint32 length field in a variable data complex type.

Fixes for each of these is in a separate commit.  Let me know if you want me to also separate them into separate pull requests.

### DIFF
--- a/main/java/uk/co/real_logic/sbe/generation/csharp/CSharpGenerator.java
+++ b/main/java/uk/co/real_logic/sbe/generation/csharp/CSharpGenerator.java
@@ -376,7 +376,7 @@ public class CSharpGenerator implements CodeGenerator
                     "        const int sizeOfLengthField = %3$d;\n" +
                     "        int limit = Limit;\n" +
                     "        _buffer.CheckLimit(limit + sizeOfLengthField);\n" +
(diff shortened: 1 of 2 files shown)


mjpt777:
Thanks, I've applied the null checks. 

If you could separate the CSharp change then that would be great. However it does feel like the CSharp code should set the type of the dataLength variable to be unsigned?

warrenfalk:
It does feel like that, but that is not feasible.  In most places, the lengths in the underlying system functions are 31 bits plus a useless sign bit (true in both C# and Java), so the real problem is that 32 bits probably shouldn't even be allowed for a length field in SBE.  However, that was too difficult to fix.  The way it is right now, the Java generator compiles but just treates it as int32, but the C# generator doesn't compile.  I decided to make the C# generated code work like the Java generated code.

mjpt777:
OK that makes sense. It is frustrating how signed and unsigned ints are not properly considered in many languages and APIs.

PR #281, opened 2015-08-10
[python] simplify handling of fixed length char arrays [behavior change]

tfgm-bud:
Get or set fixed length char arrays as a single string vs getting each character individually.

### DIFF
--- a/examples/python/decode_example.py
+++ b/examples/python/decode_example.py
@@ -19,8 +19,7 @@ print('code: '+str(car.getCode()))
 for i in range(0,car.someNumbersLength()):
     print('someNumber'+str(i)+': '+str(car.getSomeNumbers(i)))
 
-for i in range(0,car.vehicleCodeLength()):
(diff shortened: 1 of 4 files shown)


mjpt777:
Python now supported for RC3.

tfgm-bud:
@mjpt777 I'm confused by this.  Looks like python is NOT supported for RC3 at all.  Am I missing something?  -- Bud

mjpt777:
If someone is willing to submit a python port then we will be happy to accept it. @tmontgomery and I only have the time to do C++ and Java.

tfgm-bud:
@mjpt777 OK.  Your comment above said that Python was **NOW** supported.  Hence my confusion.  

If I get some time, I'll try to re implement the python code generation back into the real-logic decoder.

In the meantime, if anyone needs a python based decoder, we have open sourced our python-based SBE decoder:

  https://github.com/tfgm/sbedecoder

For performance reasons, most people would be better off with your C++ or Java decoder but it is an option.

-- Bud

mjpt777:
Typo :-)

PR #284, opened 2015-09-02
Overcome warnings in generated Java source compilation units.

benalexau:
When the SBE tool is used to generate Java source, the resulting compilation units cause warnings such as:

    The value of the field UpdateBatchDecoder.UpdatesDecoder.actingVersion is not used
    The value of the field BondSymbolInfoDecoder.parentMessage is not used

While filtering can be used to ignore such warnings within an IDE, it is preferable to overcome the warnings in a manner that does not require end user intervention.

This PR adopts the simplest possible by solution by adding ``@SuppressWarnings("all")`` to the generated compilation units. Being a ``java.lang`` annotation, this requires no additional import. It has also been available since Java 1.5. While [SuppressWarnings](http://docs.oracle.com/javase/8/docs/api/java/lang/SuppressWarnings.html) does not require a compiler to support the ``all`` warning name, compilers are required to ignore any warning names they do not recognise. Both OpenJDK and Eclipse recognise ``all``.

The cleaner alternative to this PR would be to modify ``JavaGenerator`` to only generate the fields when required. I did not pursue this path given it would require more substantial changes to ``JavaGenerator`` and I thought a more lightweight change would be preferred. If not please feel free to let me know and I can submit a more comprehensive PR.

### DIFF
--- a/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -1169,6 +1169,7 @@ public class JavaGenerator implements CodeGenerator
     private CharSequence generateClassDeclaration(final String className)
     {
         return String.format(
+            "@SuppressWarnings(\"all\")\n" +
             "public class %s\n" +
             "{\n",
             className


tmontgomery:
I'm ok with this. But would like @mjpt777 and @RichardWarburton to weigh in since they are using the Java codecs a lot right now.

mjpt777:
I'm happy with it in principle.

RichardWarburton:
I'm going to merge this patch, but I'm also going to file an issue to evaluate why the fields aren't actually used. We might be able to simplify the code generator and just not generate the dead code.

PR #293, opened 2015-09-18
Add guard to remaining __STDC_LIMIT_MACROS definitions

amluto:
10e75064f6296ca50db9d3ee17feb097169c77b3 only fixed one.  Fix the rest.

The defines can go away entirely if SBE switches to C++11.

I copied the existing style for consistency rather than using '#ifndef'


tmontgomery:
SBE C++ has moved to C++11. Thanks!

PR #294, opened 2015-09-18
Use relative includes where possible in C++ code

amluto:
Using relative includes makes it much easier to drop the shipped headers and generated code into a C++ project.  Adjust the OTF and generated headers to use relative includes.

### DIFF
--- a/main/cpp/otf_api/Field.h
+++ b/main/cpp/otf_api/Field.h
@@ -18,7 +18,7 @@
 
 #include <vector>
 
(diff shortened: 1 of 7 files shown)


amluto:
It might be worth regenerating the uk_co_real_logic_sbe_ir_generated headers on top of this.

tmontgomery:
Everything (codecs as well as OTF) should now use relative includes. Thanks, though. Closing.

PR #296, opened 2015-09-23
Adding cpp code to generator so you can retrieve the string represent…

lcs777:
…ation of an enum.


tmontgomery:
Closing this as it is out of date.

PR #344, opened 2016-04-20
[Java] default output directory of SbeTool to CWD

juddgaddie:
The docs [here](https://github.com/real-logic/simple-binary-encoding/wiki/Sbe-Tool-Guide) say the output directory defaults to current : "sbe.output.dir: Target directory for output generation, defaults to current directory."

However it attempts to write to /packagename.

Small patch to fix that.


mjpt777:
Thanks

Issue #349, opened 2016-05-21
Interface hierarchy for generated types


benalexau:
SBE Tool currently generates ``MessageEncoder`` and ``MessageDecoder`` interfaces in the ``sbe:messageSchema``-defined ``package``. The tool also ensures that flyweights generated for each ``sbe:message`` implement either ``MessageEncoder`` or ``MessageDecoder``.

I am currently developing a library for storing SBE data in the LMDB embedded key-value store (via [LMDBJNI](https://github.com/deephacks/lmdbjni)), but have encountered several difficulties in evolving this into a general-purpose library of value to others:

* The generation of interfaces in end user projects prevents them from being referenced by a general-purpose library. One alternative is to provide a code generator for end user projects, although this is a lot of extra work, not to mention inconvenient for end users to configure and use. Alternately I could model all interfaces using ``Object``, but this has adverse impacts around self-documentation, type safety and reflection overhead.
* Flyweights generated for ``sbe:composite`` types do not declare any interface. This prevents them from being referenced in interfaces. There are uses for composites as first-class citizens, for example as keys in an ordered map (particularly given the lack of versioning that a composite implies and its non-reliance on a ``messageHeader`` that would impact key ordering via its version field).
* There is no way of referencing a message or a composite as a top-level type that is independent of its encoder or decoder. This prevents symmetric interfaces that regularly need to reference both the encoder and decoder flyweights for a given type (as shown in my PoC below).

To explore what might be feasible, I have produce a [proof of concept project](https://github.com/benalexau/sbe-dao-poc). It shows:

* A [potential Agrona interface hierarchy](https://github.com/benalexau/sbe-dao-poc/tree/master/src/main/java/org/agrona/sbe) that would provide a foundation for third party libraries that need to expressively reference SBE generated types and flyweights
* A type-safe DAO interface and implementation that [illustrates use](https://github.com/benalexau/sbe-dao-poc/tree/master/src/main/java/au/com/acegi/sbedao) of the proposed Agrona hierarchy
[...]

Issue #334, opened 2016-02-26
interface for generated decoder/encoder?


andmer:
What do you think about adding an interface to the generated encoder/decoder for the common generic bits?

My interest is because I would like to pass various decoder/encoders into a some base class which contains them and hide some setup etc..

public interface Decoder<T> {
    T wrap(DirectBuffer buffer, int offset, int actingBlockLength, int actingVersion);
    etc....
}

public interface Encoder<T> {
    int sbeBlockLength();

    int sbeSchemaId();

    int sbeSchemaVersion();

    int sbeTemplateId();

    int encodedLength()

    T wrap(final MutableDirectBuffer buffer, final int offset);
   
    etc....
}

example case along the lines of some method that inits them in a baseclass that owns it:

    public void initEncoder(Encoder aEncoder) {
        mEncoderByteBuffer.clear();
        int offset = 0;
        MessageHeaderEncoder wrap = mHeaderEncoder.wrap(mEncoderUnsafeBuffer, offset);
        wrap.blockLength(aEncoderAdapter.sbeBlockLength())
                .templateId(aEncoderAdapter.sbeTemplateId())
                .schemaId(aEncoderAdapter.sbeSchemaId())
                .version(aEncoderAdapter.sbeSchemaVersion());
        offset += wrap.encodedLength();
        aEncoderAdapter.wrap(mEncoderUnsafeBuffer, offset);
    }

    public void initDecoder(Decoder aDecoder, byte[] aBytes) {
        mDecoderUnsafeBuffer.wrap(aBytes);
        int offset = 0;
        mHeaderDecoder.wrap(mDecoderUnsafeBuffer, offset);
        offset += mHeaderDecoder.encodedLength();
        aDecoder.wrap(mDecoderUnsafeBuffer, offset, mHeaderDecoder.blockLength(), mHeaderDecoder.version());
    }

mjpt777:
This would be possible and is an interesting idea. One of the goals with the move to RC3 was to have no library dependencies on SbeTool with the generated code like we did have with the old CodecUtil. 

Would you be happy with this interface being generated in the same package?

andmer:
Yeah. definitely.

mjpt777:
I'll have a look at doing something for the RC4 release.

mjpt777:
@andmer Does this match your requirements?

andmer:
Looks great, thanks Martin!

[...]

PR #350, opened 2016-05-24
Optionally generate Java interface hierarchy.

benalexau:
As per the motivations and design discussed in issue #349, this PR includes a new SBE Tool option, ``sbe.java.generate.interfaces``. This new option defaults to ``false``, but if set to ``true`` will:

* Generate an ``org.agrona.sbe`` package in the user's ``sbe.output.dir``
* Implement the new interfaces when generating the existing decoder and encoder flyweights for each ``sbe:composite`` and ``sbe:message``
* Implement the new interfaces while generating a new ``[TokenName]Structure`` class for each ``sbe:composite`` and ``sbe:message`` (these ``Structure`` classes are used for parametrization in the new interface heirarchy)

This new interface hierarchy deprecates the ``MessageEncoder`` and ``MessageDecoder`` interfaces introduced as part of issue #334. As such these interfaces are no longer generated.

Testing of this PR has been achieved as follows:

* ``GenerateFixBinaryTest`` ensures the non-generation of the interface hierarchy operates as expected
* ``JavaGeneratorTest`` ensures the generation of the interface hierarchy results in source that can successfully compile
* ``sbe-samples`` do not use the new interface hierarchy and execution of ``gradle runJavaExamples`` was successful
* Manually modifying the above ``sbe-samples`` codec generator to enable the interface hierarchy  resulted in successful generation, compilation and ``gradle runJavaExamples``
* Execution of ``gradle generateIrCodecs`` and manual review of the resulting ``sbe-tool/src/main/java/uk/co/real_logic/sbe/ir/generated`` source code reflects the expected result of ``sbe.java.generate.interfaces=false`` (and is shown in the commits below)
* Modification of the [SBE DAO POC](https://github.com/benalexau/sbe-dao-poc) to use the new SBE Tool generation was successful


mjpt777:
Thanks Ben. I'm with a client this week so it might be a little while before I can review but I'm looking forward to doing it.

mjpt777:
Other than the `MessageEncoder` and `MessageDecoder` what is the value of the other interfaces?

benalexau:
I tried to demonstrate the requirement for a fine-grained interface hierarchy in ticket #349.  The PoC linked to that ticket demonstrated my initial use case, namely a [DAO interface](https://github.com/benalexau/sbe-dao-poc/blob/master/src/main/java/au/com/acegi/sbedao/Dao.java) and [abstract DAO implementation](https://github.com/benalexau/sbe-dao-poc/blob/master/src/main/java/au/com/acegi/sbedao/AbstractDao.java) that used parametrization to contractually require an SBE composite for the key and an SBE message for the value. Being able to express this meant clients didn't need to perform casts (eg see [the test case](https://github.com/benalexau/sbe-dao-poc/blob/master/src/test/java/au/com/acegi/sbedao/user/CarDaoImplTest.java)), the abstract superclass could properly declare the abstract methods a subclass must implement, and runtime code generation (in [the example factory](https://github.com/benalexau/sbe-dao-poc/blob/master/src/main/java/au/com/acegi/sbedao/DaoFactory.java) using Agrona source compilation) could generate an implementation at runtime.

Another example is my current work-in-progress implementation of LMDB DAO code generation using a Java annotation processor (an annotation processor offers several benefits over the runtime code generation approach shown in the PoC). A portion of the annotation requires the new interfaces:

```
  /**
   * The SBE type that will be used for keys.
   *
   * @return the type of keys to store in the DAO
   */
  Class<? extends CompositeStructure> key();

  /**
   * The SBE type that will be used for values.
   *
   * @return the type of values to store in the DAO
   */
  Class<? extends MessageStructure> value();
```
Overall I've found it very useful when writing interfaces, abstract classes and annotations to be able to express exactly which type of SBE generated class is needed, and being able to enforce the relationship between encoders and decoders for a given SBE message/composite type. I hope the above links and annotation fragment demonstrate the use cases I'm trying to achieve.

mjpt777:
OK that makes sense. I can merge for now and look to extracting the interfaces to Agrona for the next release.

Issue #121, opened 2014-04-17
how to generate nested namespaces ?


chenlucan:
how to use sbe tool to generate code with nested namespace ?
I'm expecting generated cpp code like:
namespace a { namespace b { 
...
}  // namespace b
}  // namespace a

I tried the following command, it generated "namesapce a/b", which is not what I want.

java -Dsbe.target.language=cpp98 -Dsbe.target.namespace=a/b -jar sbe.jar templates.xml

Thanks

zaphoyd:
nested namespaces would be useful for us as well.

vdidenko:
See PR #351

PR #351, opened 2016-06-13
C++ nested namespaces

vdidenko:
Given a project name `proj.provider.messages` the currently generated C++ code is namespaced into `proj_provider_messages`, which is hardly an expected outcome.

This proposed PR will namespace the stubs properly nested:

    namespace proj {
    namespace provider {
    namespace messages {
    ......
    };
    };
    };

While it is in the _C++17_ standard to allow the collapsed syntax:

    namespace proj::provider::messages {
    ......
    };

It still may take some time to propagate throughout the toolchain, so my suggestion is to stick to the explicit nesting.

This PR addresses #121.

### DIFF
--- a/sbe-benchmarks/src/main/cpp/SbeCarCodecBench.h
+++ b/sbe-benchmarks/src/main/cpp/SbeCarCodecBench.h
@@ -19,7 +19,7 @@
 #include "CodecBench.h"
 #include "uk_co_real_logic_sbe_benchmarks/Car.h"
 
-using namespace uk_co_real_logic_sbe_benchmarks;
+using namespace uk::co::real_logic::sbe::benchmarks;
 
 char VEHICLE_CODE[] = {'a', 'b', 'c', 'd', 'e', 'f'};
 uint32_t SOMENUMBERS[] = { 1, 2, 3, 4, 5 };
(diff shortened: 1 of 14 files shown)


vdidenko:
Ouch, I did not notice the tests broken because I used the wrong terminal window to run the cppbuild.

Sorry.

Will be back after fixing tests.

vdidenko:
Tests adjusted, all pass on my OSX 10.11.5, Xcode 7.3.

tmontgomery:
Thanks! I like this and will review when back from traveling this week.

vdidenko:
Is there anything I should clarify to help with a review?

mjpt777:
I was chatting to @tmontgomery before the weekend and know he plans to have a look. We've been finishing off some important things on Aeron. We really appreciate the input, thanks!

tmontgomery:
Sorry for being late to this.

I love the idea of allowing for nested namespaces like this. But, we do need to be backwards compatible as this will break some existing integrations. Is it possible to make this conditional on a Java system property? The default can be to use this new nesting, though. But need a way to be able to use the old style.

vdidenko:
I think a Java system property can be pulled out into the generator - it is a Java code anyhow. I am just not a very well versed Java developer, but can try.

Which property are you thinking of?

I think it can be stored in an Ir object at construction and then used to form the namespaces array. The backward compatible version will just make a one underscore separated element namespaces array [instead of the current split by pattern](https://github.com/vdidenko/simple-binary-encoding/blob/1b82901b0ddd5860b332c76f81cd4f8a4f078e17/sbe-tool/src/main/java/uk/co/real_logic/sbe/ir/Ir.java#L69). The rest should just work, right?

vdidenko:
Added a commit which seems to do what @tmontgomery suggested. Ran a few normal and edge cases in my build, but I feel this feature needs a separate test now. Which directory/file should I look at to add a test for the property's behavior, if you feel it is needed?

tmontgomery:
I think we can be OK without a unit test for this. Normally, we should, but I am OK without one.

vdidenko:
Great, then the PR is ready to merge if there are no other improvements/changes to be made.

Issue #353, opened 2016-06-19
[C++] constexpr qualifier in static const field getters


vdidenko:
This is a follow up enhancement to the #195 issue.

I am tailoring the incoming message reaction, and it would be great to use a switch statement in the vein of the [structurechart](https://github.com/real-logic/simple-binary-encoding/issues/195#issuecomment-53376192)'s comment. Similarly to @structurechart I ran into the `constexpr` requirement.

Would it be possible to discuss a solution similar to the sample commit vdidenko/simple-binary-encoding@fe16a8ce1f1e14916d3ebe363978549a26267d5c? It introduces public constexpr fields and uses them to maintain the getter function signatures.

I have only changed the fields I _think_ I care about, but should I make the pull request similarly modifying all `static const` declarations touched by the original solution at real-logic/simple-binary-encoding@a601c48409a093abda3ad6808407712fa677dfaf?

vdidenko:
Actually, is there a specific reason, why all those values are accessible through getters, but not through namespace or class `constexpr`s? For one thing, it's simplify storing strings, as currently there is an issue with the `sbeSemanticType` in the proposed code (warning: `ISO C++11 does not allow conversion from string literal to 'char *'`)

vdidenko:
Updated the sample commit reference in the issue with a version which builds without warnings and preserves getters' signatures.

tmontgomery:
I don't think it is necessary to do all the `static const` declarations. I am a little leery of the MSVS compiler support on `constexpr` for this, but I believe it is better to go ahead. As long as the getter signatures are preserved, it should be OK for those currently using it.

vdidenko:
If the code is OK as is, then I will make it into a PR by separating it from the nested namespaces issue (constexpr support is on the top of the nested namespaces branch by accident, there is no dependency).

vdidenko:
After reviewing it, it will be easier to make this pull request after the nested namespaces work.

vdidenko:
@tmontgomery, the PR #355 is posted for a review.

vdidenko:
Solution to this issue is reasonably challenged in the PR discussion at https://github.com/real-logic/simple-binary-encoding/pull/355#issuecomment-239317154.

vdidenko:
@tmontgomery, when you refer to:

> add the new statics when it is C++11 and can be constexprs

are you talking about the functions? I thought you did not want to touch their signature at all, or do you mean to have a second function?

[...]

PR #355, opened 2016-06-29
[C++]: Static constexpr fields in CppGenerator

vdidenko:
The pull request addresses the issue #353 and is limited to HeaderStructure data in the stubs.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -145,8 +145,9 @@ public class CppGenerator implements CodeGenerator
                 generateGroups(sb, groups, BASE_INDENT);
                 out.append(sb);
                 out.append(generateVarData(className, varData, BASE_INDENT));
-
-                out.append(CppUtil.closingBraces(ir.namespaces().length) + "}\n#endif\n");
+                out.append("};\n");
+                out.append(generateStaticDefinitions(className));
+                out.append(CppUtil.closingBraces(ir.namespaces().length) + "#endif\n");
             }
         }
     }
@@ -843,6 +844,11 @@ public class CppGenerator implements CodeGenerator
             "#  include <string>\n" +
             "#  include <cstring>\n" +
             "#endif\n\n" +
+            "#if __cplusplus >= 201103L\n" +
+            "#  define SBE_CONST_KIND constexpr\n" +
+            "#else\n" +
+            "#  define SBE_CONST_KIND const\n" +
+            "#endif\n\n" +
             "#include <sbe/sbe.h>\n\n",
             String.join("_", namespaces).toUpperCase(),
             className.toUpperCase()
@@ -876,6 +882,32 @@ public class CppGenerator implements CodeGenerator
             "class %s\n" +
             "{\n",
             className
+        );
+    }
+
+    private String generateStaticDefinitions(final String className)
+    {
+        return
+            generateStaticDefinition(
+                className, "SbeBlockLength", cppTypeName(ir.headerStructure().blockLengthType()), false) +
+            generateStaticDefinition(
+                className, "SbeTemplateId", cppTypeName(ir.headerStructure().templateIdType()), false) +
+            generateStaticDefinition(
+                className, "SbeSchemaId", cppTypeName(ir.headerStructure().schemaIdType()), false) +
+            generateStaticDefinition(
+                className, "SbeSchemaVersion", cppTypeName(ir.headerStructure().schemaVersionType()), false) +
+            generateStaticDefinition(className, "SbeSemanticType", "char", true);
+    }
+
+    private static String generateStaticDefinition(
+        final String className, final String memberName, final String memberType, boolean isArray)
+    {
+        return String.format(
+            "SBE_CONST_KIND %3$s %1$s::%2$s%4$s;\n",
+            className,
+            memberName,
+            memberType,
+            (isArray ? "[]" : "")
         );
     }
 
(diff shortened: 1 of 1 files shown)


rigtorp:
Please revert this patch. It defines static member variables which causes linking issues.

vdidenko:
Can you possibly be more specific/ provide example/diagnostic? At least more context?

rigtorp:
```
SBE_CONST_KIND std::uint16_t Establish::SbeBlockLength;
SBE_CONST_KIND std::uint16_t Establish::SbeTemplateId;
SBE_CONST_KIND std::uint16_t Establish::SbeSchemaId;
SBE_CONST_KIND std::uint16_t Establish::SbeSchemaVersion;
SBE_CONST_KIND char Establish::SbeSemanticType[];
```
These needs to be in a .cpp file otherwise two .o files that include the SBE flyweight won't link due to multiple definitions of these static members.

rigtorp:
I created a GeneratedStubExample2.cpp that also includes the Car flyweight. If you try to link GeneratedStubExample.cpp and GeneratedStubExample2.cpp you get this error:
```
[ 74%] Building CXX object sbe-samples/src/main/cpp/CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o
/home/erigtorp/src/simple-binary-encoding/sbe-samples/src/main/cpp/GeneratedStubExample2.cpp: In function 'int main2(int, const char**)':
/home/erigtorp/src/simple-binary-encoding/sbe-samples/src/main/cpp/GeneratedStubExample2.cpp:31:10: warning: unused variable 'buffer' [-Wunused-variable]
     char buffer[2048];
          ^
Linking CXX executable ../../../../binaries/GeneratedStubExample
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o:(.rodata+0x0): multiple definition of `baseline::Car::SbeSemanticType'
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample.cpp.o:(.rodata+0x50): first defined here
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o: In function `main2(int, char const**)':
/home/erigtorp/src/simple-binary-encoding/sbe-samples/src/main/cpp/GeneratedStubExample2.cpp:37: multiple definition of `baseline::Car::SbeSchemaVersion'
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample.cpp.o:/home/erigtorp/src/simple-binary-encoding/generated/baseline/MessageHeader.h:48: first defined here
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o:/opt/cresearch/vendor/gcc/5.3.0/include/c++/5.3.0/iostream:74: multiple definition of `baseline::Car::SbeSchemaId'
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample.cpp.o:/home/erigtorp/src/simple-binary-encoding/generated/baseline/MessageHeader.h:48: first defined here
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o:/opt/cresearch/vendor/gcc/5.3.0/include/c++/5.3.0/iostream:74: multiple definition of `baseline::Car::SbeTemplateId'
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample.cpp.o:/home/erigtorp/src/simple-binary-encoding/generated/baseline/MessageHeader.h:48: first defined here
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o:/opt/cresearch/vendor/gcc/5.3.0/include/c++/5.3.0/iostream:74: multiple definition of `baseline::Car::SbeBlockLength'
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample.cpp.o:/home/erigtorp/src/simple-binary-encoding/sbe-samples/src/main/cpp/GeneratedStubExample.cpp:40: first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [binaries/GeneratedStubExample] Error 1
make[1]: *** [sbe-samples/src/main/cpp/CMakeFiles/GeneratedStubExample.dir/all] Error 2
make: *** [all] Error 2
```

rigtorp:
Added issue #367 to track this.

vdidenko:
It makes sense that the definitions would be guarded within one object file, yet collide if included in multiple and then linked into one. I am struggling to understand  a benefit of such diamond dependency use case.

I am sure there can be a reason (yet I am curious to learn, if you do not mind), and it aside, what may be options for keeping `SbeTemplateId` available for use in a `switch` statement?

Considering that changing `static const std::uint16_t sbeTemplateId(void)` signature is not an option one way would be to have a second, `constexpr` static member function returning `SbeTemplateId` in addition to the existing const function. A little messy, But is the only thing which comes to mind immediately.

Any other thoughts?

vdidenko:
Instead of hopping between the issues, would it be OK to reopen the #353? My preference is to keep reasoning in one place.

rigtorp:
This is one solution:
```
struct Flyweight {
  enum {
    SbeTemplateId = x,
    // etc
  };
 // rest
}
```

rigtorp:
> It makes sense that the definitions would be guarded within one object file, yet collide if included in multiple and then linked into one. I am struggling to understand a benefit of such diamond dependency use case.

It's not a diamond dependency. I just have two parts of a program using the same flyweight. With the new codegen that is not possible.

tmontgomery:
Another option might be to only add the new statics when it is C++11 and can be constexprs.

vdidenko:
Ok, sorry, long day. Makes sense.

rigtorp:
> Considering that changing static const std::uint16_t sbeTemplateId(void) signature is not an option one way would be to have a second, constexpr static member function returning SbeTemplateId in addition to the existing const function. A little messy, But is the only thing which comes to mind immediately.

`static constexpr uint16_t sbeTemplateId()`  works fine to use in switch() with gcc-5.3.0 -std=c++14. Maybe others too.

rigtorp:
> Another option might be to only add the new statics when it is C++11 and can be constexprs.

Just add `constexpr` to the current functions. Should work on C++11.

PR #360, opened 2016-07-06
[Java] Rationalize display format

xdaragon:
Use the following conventions:
* () for sub-messages (composites, group instances)
* [,] for arrays and repeating group
* {,} for sets


xdaragon:
Added the enum for clarity sake, but not convinced it is successful, external opinion welcome.

Issue #367, opened 2016-08-11
[C++] New codegen causes linking issues


rigtorp:
Any two .cpp files including the same SBE flyweight won't link.

How to reproduce:
Apply this patch:
```
diff --git a/sbe-samples/src/main/cpp/CMakeLists.txt b/sbe-samples/src/main/cpp/CMakeLists.txt
index 76bee3f..31fbd49 100644
--- a/sbe-samples/src/main/cpp/CMakeLists.txt
+++ b/sbe-samples/src/main/cpp/CMakeLists.txt
@@ -29,7 +29,7 @@ add_custom_command(
 )
 add_custom_target(examples_codecs DEPENDS ${GENERATED_CODECS})
 
-add_executable(GeneratedStubExample GeneratedStubExample.cpp)
+add_executable(GeneratedStubExample GeneratedStubExample.cpp GeneratedStubExample2.cpp)
 add_executable(OtfExample OtfExample.cpp)
 target_link_libraries(OtfExample sbe)
 
diff --git a/sbe-samples/src/main/cpp/GeneratedStubExample2.cpp b/sbe-samples/src/main/cpp/GeneratedStubExample2.cpp
new file mode 100644
index 0000000..2153c81
--- /dev/null
+++ b/sbe-samples/src/main/cpp/GeneratedStubExample2.cpp
@@ -0,0 +1,3 @@
+#include "baseline/MessageHeader.h"
+#include "baseline/Car.h"
+
```

Try and build the C++ samples:
```
[ 29%] Built target gtest
[ 33%] Built target sbe
[ 37%] Built target codecs
[ 40%] Built target BoundsCheckTest
[ 44%] Built target CodeGenTest
[ 48%] Built target CompositeElementsTest
[ 51%] Built target CompositeOffsetsCodeGenTest
[ 55%] Built target CompositeOffsetsIrTest
[ 59%] Built target GroupWithDataTest
[ 62%] Built target MessageBlockLengthTest
[ 66%] Built target Rc3OtfFullIrTest
[ 70%] Built target examples_codecs
Linking CXX executable ../../../../binaries/GeneratedStubExample
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample2.cpp.o:(.rodata+0x0): multiple definition of `baseline::Car::SbeSemanticType'
CMakeFiles/GeneratedStubExample.dir/GeneratedStubExample.cpp.o:(.rodata+0x50): first defined here
[...]

PR #369, opened 2016-08-15
Fix #353 #355 #366

### LINKED ISSUE #353: [C++] constexpr qualifier in static const field getters

vdidenko:
Address issue #353, PRs #355, #366
Regenerate IR stubs using the fixed constexpr generator

### DIFF
--- a/sbe-tool/src/main/cpp/uk_co_real_logic_sbe_ir_generated/ByteOrderCodec.h
+++ b/sbe-tool/src/main/cpp/uk_co_real_logic_sbe_ir_generated/ByteOrderCodec.h
@@ -22,9 +22,9 @@
 #endif
 
 #if __cplusplus >= 201103L
-#  define SBE_CONST_KIND constexpr
+#  define SBE_CONSTEXPR constexpr
 #else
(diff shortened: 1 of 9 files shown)


vdidenko:
Also addresses #367 (dup of #353)

rigtorp:
:+1: Looks good to me

PR #370, opened 2016-08-15
Correctly encode int64 minimum value.

rigtorp:
Due to a quirk in C++ int64 minimum value can not be encoded as
literal. You need to use INT64_MIN constant.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -1779,6 +1779,10 @@ public class CppGenerator implements CodeGenerator
 
             case INT64:
                 literal = value + "L";
+                if (value.equals("-9223372036854775808"))
+                {
+                    literal = "INT64_MIN";
+                }
                 break;
 
             case UINT64:


rigtorp:
@tmontgomery This schema ftp://ftp.moex.com/pub/FORTS/TWIME/doc/wire.xml generates code that causes warnings due to "-9223372036854775808L" literal overflowing the int64. This is a fix for that.

rigtorp:
This is what GCC 6.1 will give you:
```
1 : warning: integer constant is so large that it is unsigned
int64_t i = -9223372036854775808L;
```

tmontgomery:
Unfortunately, the FIX/SBE spec sets this value. https://github.com/FIXTradingCommunity/fix-simple-binary-encoding/blob/master/v1-0-RC4/doc/02FieldEncoding.md#range-attributes-for-integer-fields

rigtorp:
That's fine. `INT64_MIN` is defined to equal -9223372036854775808. It's just a quirk with the way C/C++ is parsed that causes this.

rigtorp:
http://stackoverflow.com/questions/11270084/why-are-there-differing-definitions-of-int64-min-and-why-do-they-behave-differe#11270104

http://en.cppreference.com/w/cpp/types/integer

rigtorp:
@tmontgomery Just so you understand what's going on here:

This compiles:
```
#include <cstdint>
static_assert(INT64_MIN == -9223372036854775808L);
```
With GCC 6.1 output:
```
2 : warning: integer constant is so large that it is unsigned
static_assert(INT64_MIN == -9223372036854775808L);
^~~~~~~~~~~~~~~~~~~~
Compiled ok
```

With clang 3.8 output:
```
2 : warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
static_assert(INT64_MIN == -9223372036854775808L,"");
^
1 warning generated.
Compiled ok
```

`INT64_MIN` is defined as:
```
# define INT64_MIN              (-__INT64_C(9223372036854775807)-1)
```
Conclusion:
`INT64_MIN` is equal to -9223372036854775808, which is the smallest 64bit signed integer. 
`INT64_MIN` is defined with a trick to avoid the above warning.

It would be great to fix this warning.

PR #376, opened 2016-08-25
[C++]: make constant fields static constexpr

abbradar:
Constant properties now are truly constant, which allows us to perform compile-time usage of them.


abbradar:
Ouch, this is broken -- I've cherry-picked the wrong commit. Closing till I fix this.

PR #384, opened 2016-09-01
Enable extra warnings (-Wextra)

rigtorp:
Enable extra warnings and fix those warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ set(GTEST_BINARY_DIR ${binary_dir})
 find_package(Threads)
 
 if(UNIX)
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fexceptions -g -Wall -Wpedantic")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fexceptions -g -Wall -Wpedantic -Wextra -Wno-unused-parameter")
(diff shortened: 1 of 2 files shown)


rigtorp:
@tmontgomery What do you think about this patch?

config/suppress.xml
```diff
@@ -4,4 +4,5 @@
         "http://www.puppycrawl.com/dtds/suppressions_1_0.dtd">
 <suppressions>
     <suppress files=".*generated.*" checks="."/>
+    <suppress files="CppGenerator.java" checks="MethodLength"/>
```
tmontgomery: This is not something we want to suppress.

rigtorp: I don't want to split up the constructor codegen function for no good reason. Should I do something like this:
```
void generateConstructor() {
  generatePart1();
  generatePart2();
}
```

mjpt777: Yes it is best to break up large methods into logical blocks so they can be reasoned about.

ksergey:
a looking forward 80c53b6 in next release

rigtorp:
@tmontgomery @mjpt777 I've updated this PR to ignore -Wtype-limits in wrapForEncode instead of the more complicated change i previously had.

This PR is a superset of #421

rigtorp:
@tmontgomery @mjpt777 Any thoughts on this?

tmontgomery:
Sorry this took so long. Thanks!

PR #385, opened 2016-09-07
[C++]: Expose buffer in fixed flyweights

abbradar:
We already expose them for messages, so this shouldn't pose problems.


sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
```diff
@@ -1299,6 +1299,14 @@ private static CharSequence generateFixedFlyweightCode(final String className, f
             "    static SBE_CONSTEXPR const std::uint64_t encodedLength(void)\n" +
             "    {\n" +
             "        return %2$s;\n" +
+            "    }\n\n" +
+            "    std::uint64_t offset(void) const\n" +
```
rigtorp: Should this be `size_t` instead?

abbradar: I don't think so -- SBE uses `uint64_t` internally for offsets everywhere (this is wrong, but that's another story).

PR #391, opened 2016-09-20
[C++]: Don't generate copy and move constructors and operators

abbradar:
They are not needed because they are generated automatically as long as you don't have destructors and just copying all the fields should be the correct way. They are also invalid as they are because they can't copy an ampty flyweight (one that was constructed with a default constructor). This happens because we try to call `reset` which validates buffer sizes.


tamboril:
This broke my code.  I think it's because without the call to reset(),  `m_positionPtr` is copied as a pointer into the original object.

PR #397, opened 2016-10-16
[Java] add new convenience methods to get char arrays as string.

Starch:
Convenience methods for extracting char arrays as String in decoders (and also the symmetric setter in encoders). Declared encoding is used and zero bytes are removed.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -1621,6 +1621,23 @@ public class JavaGenerator implements CodeGenerator
                 fieldLength,
                 generateArrayFieldNotPresentCondition(token.version(), indent),
                 offset));
+
+            sb.append(String.format(
+                "\n" +
+                indent + "    public String %s()\n" +
+                indent + "    {\n" +
+                "%s" +
+                indent + "        final byte[] dst = new byte[%d];\n" +
+                indent + "        buffer.getBytes(this.offset + %d, dst, 0, %d);\n\n" +
+                indent + "        int end = 0;\n" +
+                indent + "        for (; end < %d && dst[end] != 0; ++end);\n\n" +
+                indent + "        return new String(dst, 0, end, %s);\n" +
+                indent + "    }\n\n",
+                formatPropertyName(propertyName),
+                generateStringNotPresentCondition(token.version(), indent),
+                fieldLength, offset,
+                fieldLength, fieldLength,
+                charset(encoding.characterEncoding())));
         }
 
         return sb;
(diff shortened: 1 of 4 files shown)


mjpt777:
We are trying to have the SBE generated stubs not have any dependencies on SBE jars, only Agrona. Can you remove the dependency or migrate the necessary code to Agrona?

Starch:
Unless I am mistaken, there is no dependency to any ``uk.co.real_login.sbe`` code in *generated* code. The code added in ``uk.co.real_logic.sbe.generation.java.JavaUtil`` is only used during generation.

Here are samples of generated code for both UTF-8 (present in StandardCharsets) and ISO-8859-15 (not present in StandardCharsets):
 * template
```xml
    <type name="char18" length="18" primitiveType="char" characterEncoding="ISO-8859-15"/>
    <type name="char102" length="102" primitiveType="char"/>
 * Encoder
```java
    public String fullInstrumentName()
    {
        final byte[] dst = new byte[102];
        buffer.getBytes(this.offset + 14, dst, 0, 102);

        int end = 0;
        for (; end < 102 && dst[end] != 0; ++end);

        return new String(dst, 0, end, java.nio.charset.StandardCharsets.UTF_8);
    }
    public String instrumentName()
    {
        final byte[] dst = new byte[18];
        buffer.getBytes(this.offset + 116, dst, 0, 18);

        int end = 0;
        for (; end < 18 && dst[end] != 0; ++end);

        return new String(dst, 0, end, java.nio.charset.Charset.forName("ISO-8859-15"));
    }
```

 * Decoder
```java
   public StandingDataEncoder fullInstrumentName(final String src)
    {
        final int length = 102;
        final byte[] bytes = src.getBytes(java.nio.charset.StandardCharsets.UTF_8);
        if (bytes.length > length)
        {
            throw new IndexOutOfBoundsException("string too large for copy: byte length=" + bytes.length);
        }

        buffer.putBytes(this.offset + 14, bytes, 0, bytes.length);

        for (int start = bytes.length; start < length; ++start)
        {
            buffer.putByte(this.offset + 14 + start, (byte) 0);
        }

        return this;
    }

    public StandingDataEncoder instrumentName(final String src)
    {
        final int length = 18;
        final byte[] bytes = src.getBytes(java.nio.charset.Charset.forName("ISO-8859-15"));
        if (bytes.length > length)
        {
            throw new IndexOutOfBoundsException("string too large for copy: byte length=" + bytes.length);
        }

        buffer.putBytes(this.offset + 116, bytes, 0, bytes.length);

        for (int start = bytes.length; start < length; ++start)
        {
            buffer.putByte(this.offset + 116 + start, (byte) 0);
        }

        return this;
    }
```

mjpt777:
Sorry I mistook the used to JavaUtil to be a dependency in the generated stub.

Starch:
thanks!

mjpt777:
Thanks for the contribution :)

PR #403, opened 2016-10-31
[C++]: Add rewind() function to reset flyweight to the message start

abbradar:
It's useful when you run a two-pass algorithm on the data from message, and don't want to re-create a flyweight (or copy data).

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -1388,6 +1388,10 @@ public class CppGenerator implements CodeGenerator
             "    {\n" +
             "        return m_buffer;\n" +
             "    }\n\n" +
+            "    void rewind(void)\n" +
+            "    {\n" +
+            "        m_position = m_offset + m_actingBlockLength;\n" +
+            "    }\n\n" +
             "    std::uint64_t actingVersion(void) const\n" +
             "    {\n" +
             "        return m_actingVersion;\n" +


mjpt777:
This is not sufficient if the message has repeating groups or var data. You'd need to store the initial offset and reset it on rewind.

abbradar:
Hm, I'm unsure in my understanding of how SBE works, so please bear with me: as far as I see, `position` is the only field that is mutated by group flyweights. It's implemented by passing `positionPtr` to group flyweight that can increment it while it moves through the group (via `next()` calls). Then, the next group can be read because `position` now points to the end of previous group. Therefore, resetting `position` is sufficient, as `m_offset` and other fields stay constant after construction.

tmontgomery:
Why not just use `wrapForDecode()`? It is meant for reuse of the flyweight. While you would still need to keep the buffer and actingBlockLength, etc. I would think a 2 pass algorithm would have that kept around anyway.

abbradar:
First it would somewhat break architecture, because the actual algorithm itself doesn't need to know about pointers, sizes and stuff. It needs the message, and making it to wrap (and re-wrap later) flyweight by itself is mixing responsibilities.

Second, rewinding via this call is a little bit faster because `reset()` sets additional fields (which are not actually changed) and checks sizes (via `position()`, if enabled). We don't actually need this at the point we want to reset, because all fields are already there and checked.

tmontgomery:
WRT responsibilities, a flyweight depends on buffer, offset, acting version, and acting blocklength. While I agree the algorithm shouldn't care about those, a subclass of the codec could be created that holds those dependencies and adds a `rewind` functionality using a wrapForDecode. You could even use a lambda to hold the dependencies if you wish so that the algo doesn't hold it. Many ways to keep those responsibilities separated.

You are correct in that the current version doesn't need to set some of the values. But I am unsure if rewind is the right place to optimize that behavior.

If you can add _new_ code gen and bounds check tests that exercise rewind so that when we make changes that effect parsing (which we would like to do) we have some assurance that rewind works for all the normal cases, then I would be OK with adding it. To be clear, rewind would need to be exercised on all access paths for the codec and not just end of parse.

abbradar:
> You are correct in that the current version doesn't need to set some of the values. But I am unsure if rewind is the right place to optimize that behavior.

I actually don't think that `reset()` can do less than it does now -- it needs to validate size that it was passed and fill other fields. It's because when `rewind()` is called all this validation was already finished we can skip it and just return `m_position` to the original value. I don't think we can do this in any other currently existing call.

> If you can add new code gen and bounds check tests that exercise rewind so that when we make changes that effect parsing (which we would like to do) we have some assurance that rewind works for all the normal cases, then I would be OK with adding it. To be clear, rewind would need to be exercised on all access paths for the codec and not just end of parse.

I'll look at adding tests for this -- this seems a good idea to me.

mjpt777:
I'm closing due to lack of activity.

PR #404, opened 2016-10-31
[C++] Add Position and Size values

abbradar:
Needed for low-level operations on message buffers, when you want to operate on a part of message representation in memory. For example, this can be used to update changed parts of a message in an external hardware buffer.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -1009,6 +1009,15 @@ public class CppGenerator implements CodeGenerator
             propertyName,
             generateLiteral(primitiveType, token.encoding().applicableMaxValue().toString())));
 
+        sb.append(String.format(
+            "\n" +
+            indent + "    static SBE_CONSTEXPR const std::size_t %1$sEncodingLength(void)\n" +
+            indent + "    {\n" +
+            indent + "        return %2$d;\n" +
+            indent + "    }\n",
+            propertyName,
+            token.encoding().primitiveType().size()));
+
         return sb;
     }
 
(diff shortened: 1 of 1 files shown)


mjpt777:
For position have you considered how this would work with var data or repeating groups?

abbradar:
No, this wouldn't work with those types because we don't (obviously) know positions for them until runtime. My use-case is fixed fields -- for variable case one can use `position()` to achieve similar effect.

mjpt777:
This would confuse others unless the solution is more general purpose. Please consider this otherwise the this is not practical.

abbradar:
The problem with groups/var data is that we can't determine their position at compile time (read: `constexpr`), so certain optimizations can't be performed (like calculation of total offset and size of a memory chunk to transmit at compile time). I don't see any way to implement this for non-fixed fields while maintaining this property -- maybe I'm missing some idea?

tmontgomery:
For C++, the generated encodedLength() method can be used to get the length. It accounts for var data and repeating groups. If those types are not present, then it is the block length. You can also return the generated sbeBlockLength() (constexpr when possible) and use that as well.

abbradar:
Unfortunately, `encodedLength()` is not sufficient, because:
1) There's no way to get position of specific fixed fields;
2) Similarly, we can't find out size of a field.

My usecase, as have been mentioned before, is to use that information to push only updated chunks of the message to an external device buffer. For example, I have a message in that buffer that I've sent previously and I want to re-send it, but change only one value. Instead of pushing the whole message anew, I can update this field in my local in-memory copy of that message and push only the changed part (several bytes for an integral field).

tmontgomery:
I see. The use of terms here is very misleading. Please use different terms that are more expressive than `position` and `size`. Position has a very different connotation for the codecs.

Some suggestions ( @mjpt777 should weigh in on them).
1. `NAMEFieldOffset` for position
1. `NAMEFieldLength` for size

Field isn't right either, but I can't think off hand of a better description. Very open to better suggestions.

mjpt777:
The SBE spec refers to "encoding" rather than field, as do we in the IR.

tmontgomery:
Encoding makes send to me.

`NAMEEncodingOffset` and `NAMEEncodingLength`

abbradar:
@tmontgomery Good idea! Renamed.

mjpt777:
I've committed a Java equivalent. See https://github.com/real-logic/simple-binary-encoding/commit/73de4f77d22cf9a8b3b30583760a2a67d91529c9

PR #406, opened 2016-11-02
change to use std::numeric_limits min & max

derekxgl:


tmontgomery:
My concern is that the numeric_limits do not seem to define specializations for the fixed width types. At least consistently across clang, gcc, and MSVC versions. Nor even the spec. I've run into issues in the past with this. Also, there is the issue of the NULL definitions for FIX/SBE fields and how they map to those.

So, let me dig into this when I have time.

derekxgl:
INT8_MIN, etc are somehow not defined in g++ 5.2 on my RedHat 6 build. For gcc, int8_t is essentially signed char. int16_t is short. All gets defined in stdint.h. <limits> has specialization for all signed, and unsigned char, short, int, long long. Think it's likely all recent compilers should support numeric_limits on fixed width types (http://en.cppreference.com/w/cpp/types/numeric_limits ).

derekxgl:
`int8_t` etc are introduced in c++11. It's compiler's bug if `numeric_limits` does not give correct value. I tested g++ and clang. Both work fine.

One caveat is that `int8_t` is actually `signed char`, `std::numeric_limits<int8_t>::min()` returns -128 of type `signed char`. If `std::cout `directly, -128 of `signed char` is not printable character. `INT8_MIN` is simply -128, which is of type `int`. So no `std::count` issue.

daveshep666:
I'm having compiler errors with this change on Visual Studio 2015 (other versions of Visual Studio may have the same issue). It's because min and max are defined as function macros and are being expanded with the calls to min and max on numeric_limits.

one possible fix, although not very neat, would be to include an extra set of parenthesis that prevents that expansion, for example:
#define SBE_NULLVALUE_INT8 (std::numeric_limits<std::int8_t>::min)()

PR #419, opened 2016-12-12
Add golang generation support (snapshot 1) 

billsegall:
[MarketFactory](https://www.marketfactory.com) are currently using sbe-tool for Java and C++ and would like to extend our usage to include golang.  We are happy to maintain a golang backend for sbe-tool and developing one is underway.

I thought that in in the interest of starting a conversation that an initial pull request would help us understand how you like to manage accepting contributions. This pull request has been squashed to two commits.

The first commit results from golang's use of 'make' as a predeclared identifier requiring it to be added to the list of keywords.

The second commit is sufficient to get the car example-schema working but should be regarded as a first snapshot during development.

Further development steps (as currently imagined) are contained in gocode/README.md.

I understand that you will probably have many comments on this snapshot and I'll be very happy to refine things with your input.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ build-local.properties
 deps
 /bin
 build
+classes
 .gradle
 target
 GTAGS
@@ -30,7 +31,7 @@ CMakeFiles
 lib
 gtest-prefix
 CMakeCache.txt
-Makefile
+./Makefile
 binaries
 codecs
 Testing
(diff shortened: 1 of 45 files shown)


mjpt777:
Thanks @billsegall. I'm travelling this week. I'll have a look when I return home.

mjpt777:
I'm happy to merge this if you are willing to provide ongoing support?

One thing I noticed was the rename of "make" to "brand" in the examples. I've worked in the automotive industry and people tend to talk of make, model, and derivative to identify a vehicle.  The make and model are both brands, e.g. Ford and Focus are both brands. Any particular reason for this?

billsegall:
We are definitely able to provide ongoing support. 

The make to brand change results from 'make' being a reserved word in Golang. I couldn't come up with a better alternative but I understood this was possibly contentious which is why I made it  a separate pull request.

mjpt777:
OK that makes sense on the the "make" change. I did not realise it was a keyword in go. Maybe we name it "manufacturer"?

billsegall:
I'd be happy to switch to manufacturer. I'm away until Dec 27 and can resubmit when I return.

PR #421, opened 2016-12-19
[C++]: Remove const qualifiers where they have no effect

abbradar:
This qualifier should have no effect because you are returning a copy.


mjpt777:
@tmontgomery Do you have a view on that making the next release?

rigtorp:
Also so see #384 

I recommend to enable warning -Wignored-const-qualifiers with this patch so we prevent this in the future.

tmontgomery:
@rigtorp is this still relevant given #384?

rigtorp:
@tmontgomery No, #384 is a super set of this and will warn about these issues if reintroduced in the future.

tmontgomery:
Closing this. Per @rigtorp

PR #426, opened 2017-01-17
golang backend phase2

billsegall:
This release is mostly functionally complete but see gocode/README.md for next steps. Note, there are API Changes and added methods.

Handle extension logic so that schemas may be extended.

Implement Deprecated().

General code improvements/refactoring.

Vastly improve group handling code.

Add helper methods to choices.

Add much more testing with bug fixes.

Properly introduce range checking on encode/decode so we can better check for correctness.

Improve the documentation/design doc.

Improve the build environment (but see README.md)

Add some initial benchmark cases before beginning performance work.

Fix go fmt - generated code is now clean.


mjpt777:
Thanks @billsegall

PR #427, opened 2017-01-28
Make the byte order available in the IR

billsegall:
This doesn't push the byteOrder down into each Message as I think it belongs with the message schema beside package, id, version, and semanticVersion.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/ir/IrDecoder.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/ir/IrDecoder.java
@@ -25,6 +25,7 @@ import uk.co.real_logic.sbe.ir.generated.TokenCodecDecoder;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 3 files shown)


sbe-tool/src/main/java/uk/co/real_logic/sbe/ir/IrDecoder.java
```diff
@@ -97,7 +99,7 @@ public Ir decode()
             i = captureHeader(tokens, 0);
         }
 
-        final Ir ir = new Ir(irPackageName, irNamespaceName, irId, irVersion, semanticVersion, irHeader);
+        final Ir ir = new Ir(irPackageName, irNamespaceName, irId, irVersion, semanticVersion, byteOrder, irHeader);
```
mjpt777: What if the existing encoded IR was in BIG_ENDIAN format? Will this not result in a bug?

mjpt777:
Has been addressed with this commit. https://github.com/real-logic/simple-binary-encoding/commit/0ab08e0840f5c4431df2f521c4cf33d8d3eb2c25

Issue #324, opened 2016-01-23
Getting started with SBE.


Lee0112358:
New to SBE, would like to start learning more. I have recently succesfully built and installed SBE. I am using C++ on linux. Where exactly is the sbe-tool located in the root folder? I have an example-schema.xml schema I would like to start using but how does one actually create the c++ stubs? 

Thanks in advance.

tmontgomery:
You will want to run the Java application, SbeTool, and set the property `sbe.target.language` to `Cpp`. That will generate C++ stubs for the schema you input.

Lee0112358:
Thanks for the response. As of now, I am using the following command:
java -Dsbe.target.language=cpp98 -jar sbe-1.3.1-RC3-SNAPSHOT.jar example-schema.xml. 

The sbe-1.3.1-RC3-SNAPSHOT.jar I am running is located in /simple-binary-encoding-master/build/libs.

Getting the following error. 
no main manifest attribute, in sbe-1.3.1-RC3-SNAPSHOT.jar

Do I have the correct sbe-tool path? There are a bunch of other directories created when I installed SBE and not sure which .jar to use.

mjpt777:
If I check out the latest and run the below it works for me in the project dir. Alternatively you could fetch the released JAR from Maven Central. 

http://search.maven.org/#search%7Cga%7C1%7Csbe

```
java -Dsbe.target.language=cpp \
    -Dsbe.output.dir=target \
    -jar sbe-tool/build/libs/sbe-tool-1.3.2-RC3-SNAPSHOT-all.jar \
    sbe-samples/src/main/resources/example-schema.xml
```

mjpt777:
https://github.com/real-logic/simple-binary-encoding/wiki/Sbe-Tool-Guide

Lee0112358:
That worked, thanks very much for the help. Appreciate the time.

dxwu88:
If I want to generate C# code, how should I specify the language option sbe.target.language? Thanks!

mjpt777:
@dxwu88 https://github.com/real-logic/simple-binary-encoding/blob/master/build.gradle#L674

dxwu88:
It works. Thanks!!!

________________________________

PR #466, opened 2017-06-11
Explain the produced Jars to avoid missuses

olibre:
I spent some hours to understand that I was trying to execute the wrong Jar:  
I was wondering for hours what was wrong in the building process about produced `build/libs/sbe-all-1.7.0.jar` without `META-INF/MANIFEST.MF`... 😞 

I think I am not alone because it was also the issue #324

Instead of creating a new issue to request someone to document that point, I propose in this pull reuest to provide some documentation in the README.md about produced Jars, and also a warning to prevent using the wrong Jars.

It is important to keep the error message *"no main manifest attribute"* within the README.md in case someone is searching this error on the GitHub project or in the web (e.g. googling *"error SBE no main manifest attribute"*). With some luck this person will helpfully find this information within the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,9 +83,18 @@ Run the Java examples
 
 Distribution
 ------------
-Jars for the executable, source, and javadoc for the various modules can be found in
+Jars for the executable, source, and javadoc for the various modules can be found in the following directories:
 
-    <module>/build/libs
+    sbe-benchmarks/build/libs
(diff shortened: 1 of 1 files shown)


mjpt777:
Thanks

mjpt777:
I've updated the build so that jars are not produced for the root project which should help for future users.

PR #475, opened 2017-06-27
Issue472 Handle presence="optional"

billsegall:
Add intialization to NullValue for optional fields.

Allow NullValue for optional fields in RangeCheck.

Add test case.


mjpt777:
Thanks

meyerdan:
Thanks for fixing this so quickly! 💯

billsegall:
Thanks for the most excellent bug report. The golang backend is quite new so there will be lurking issues.

Issue #368, opened 2016-08-13
[C++] Add noexcept specifiers where it's apropriate 


rigtorp:
Setters and getters without bounds checking should be marked `noexcept`.

tmontgomery:
Agreed.

PR #480, opened 2017-07-15
[C++] Add noexcept specifier

fracek:
See #368 

You can see the generate files [here](https://gist.github.com/fracek/82c66357e039eebd09281c562e9cfd29), if I missed anything let me know!


tmontgomery:
This may not work on VS2013 and possibly some older compilers. However, I think this is a good idea anyway to add. And if it needs tweaked, we can do that.

PR #510, opened 2017-09-05
Move the static generateLiteral(...) method from JavaGenerator to JavaUtil and make it public

l-yara:
The `JavaUtil` class contains a lot of useful methods for creating the customer-specific SBE-based tools . The `generateLiteral(final PrimitiveType type, final String value)` method is a nice addition to this toolbox.

NOTE: any objection to moving the `JavaGenerator.eachField(final List<Token> tokens, final BiConsumer<Token, Token> consumer)` method to the `JavaUtil` too? It is turned public in the current SNAPSHOT so somebody uses it already but it does not feel right to use `JavaGenerator` for utility methods.


mjpt777:
Thanks. I've extracted the `eachField` method with this commit. https://github.com/real-logic/simple-binary-encoding/commit/daad506d2c7a39ac094502cc920e92a879cfb5a2

Issue #517, opened 2017-10-02
C++: char arrays potentially read/write too far


MatthiasKauer:
Hi,
I have observed the following behavior.

Using a char array field like the one from the baseline example (length increased to 20 so it is above initial capacity of std::string on my platform):

    <type name="VehicleCode" primitiveType="char" length="20" characterEncoding="ASCII"/>

```
//main.cpp
#include <iostream>
#include <string>

#include "baseline/Car.h"
#include "baseline/MessageHeader.h"

int main() {
  using namespace baseline;

  std::cout << "hi" << std::endl;

  char buffer[2048];
  MessageHeader hdr;
  Car car;

  car.wrapForEncode(buffer, hdr.encodedLength(), sizeof(buffer));
  car.putVehicleCode("ab");
  std::cout << car.getVehicleCodeAsString() << std::endl;
  std::cout << car.vehicleCode() << std::endl;

  std::string s = "SOMEVALUES";

  std::cout << "bye" << std::endl;
}
```

Output:
```
o A2F6D matt@hp840:~/dvl/sbe_bug $ g++ -g -O2 -I ~/dvl/sbe_bug -std=c++14 main.cpp && ./a.out 
hi
abSOMEVALUESbye
ab
bye
```

Valgrind also finds errors. The problem seems to be in the following generated code:
```

    Car &putVehicleCode(const char *src)
    {
        std::memcpy(m_buffer + m_offset + 32, src, sizeof(char) * 20);
        return *this;
    }

    std::string getVehicleCodeAsString() const
    {
        std::string result(m_buffer + m_offset + 32, 20);
        return result;
    }

    Car &putVehicleCode(const std::string& str)
    {
        std::memcpy(m_buffer + m_offset + 32, str.c_str(), 20);
        return *this;
    }
```
The length of the string could be checked. I am not sure what the best approach is for the `char *` setter.

Are you interested in a pull request for this?

Best regards,
Matthias

tmontgomery:
Did you turn off bounds checking?

tmontgomery:
Also, those setters assume the app is checking termination. It is only protecting the underlying buffer.

Also, the field is a set length in size. It's 20 bytes. If the app passes less than that, it is the apps fault that it didn't adhere to the format of the field.

MatthiasKauer:
I believe bounds checking is on by default and I left it that way. My compilation options are quite short and included above.

[...]

Issue #506, opened 2017-08-30
[C++] Undefined behaviour in generated code


rigtorp:
The C++ code generator typically generates code like this for accessing integer fields:
```cpp
    std::int32_t securityID(void) const
    {
        return SBE_LITTLE_ENDIAN_ENCODE_32(*((std::int32_t *)(m_buffer + m_offset + 24)));
    }

    NewOrderSingle &securityID(const std::int32_t value)
    {
        *((std::int32_t *)(m_buffer + m_offset + 24)) = SBE_LITTLE_ENDIAN_ENCODE_32(value);
        return *this;
    }
```

The compiler assumes that a pointer to a int32_t has the correct alignment. With this generated code the alignment requirement might not fulfilled. On amd64 this is fine as long as the compiler doesn't try to use SSE or AVX instructions, but it's not safe in general to assume it will work.

The solution is to use memcpy like this (https://chromium.googlesource.com/chromium/src.git/+/master/base/bit_cast.h):
```cpp
template <class Dest, class Source>
inline Dest bit_cast(const Source& source) {
   static_assert(sizeof(Dest) == sizeof(Source),
   "bit_cast requires source and destination to be the same size");
   static_assert(base::is_trivially_copyable<Dest>::value,
   "bit_cast requires the destination type to be copyable");
   static_assert(base::is_trivially_copyable<Source>::value,
   "bit_cast requires the source type to be copyable");
    
   Dest dest;
   memcpy(&dest, &source, sizeof(dest));
   return dest;
}
```
This should optimize to a single load load on amd64, guaranteed not to use instructions requiring alignment.

mjpt777:
The schema, as per the SBE specification, can use the `offset` attribute for aligning fields when targeting platforms which do no support unaligned access.

rigtorp:
Currently our unit tests do not pass due to this issue and we had to disable alignment check in ubsan. By using the memcpy trick we can inform the compiler that the load is unaligned and make sure it will not use instructions that require alignment. On amd64 the resulting binary should be the same, except the code is guaranteed not to break if some new optimization pass is introduced that would generate SSE or AVX instructions for this code.

mjpt777:
Have you used the `offset` attribute to align you fields?

rigtorp:
I don't control the specification, I'm using what is provided by the exchange etc. But that is not the problem. 

[...]

PR #522, opened 2017-10-21
[C++] Fix String Handling

marayl:
This patch fixes several string handling issues in the current C++ code-generator.
The following Gist describes these issues in detail:

https://gist.github.com/marayl/729a6305335e547acc927a0428d77e25

See related issue #517.


sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
```diff
@@ -1121,51 +1121,83 @@ private CharSequence generateArrayProperty(
...
+        }
+        else
         {
             sb.append(String.format(
+                indent + "    std::uint64_t get%1$s(char *dst, std::uint64_t length) const SBE_NOEXCEPT\n" +
+                indent + "    {\n" +
+                indent + "        std::uint64_t minlen = std::min<std::uint64_t>(length, %3$d);\n" +
+                indent + "        const char* src = m_buffer + m_offset + %2$d;\n" +
+                indent + "        const char* end = sbe_util::stpncpy(dst, src, minlen);\n" +
+                indent + "        if (length > 6)\n" +
```
dpwegener: Is the hard coded value of 6 here intentional?  If the intention is to null terminate dst when possible, I would expect that the compare would be against minlen.

marayl: You're absolutely right. Well spotted! Pushing update now.

dpwegener: Glad to help.  Getting this accepted would help overcome the issues described your Gist.

marayl:
This PR has been open for some time now. Is there anything I can do to help move this forward?

tmontgomery:
Sorry for the lack of feedback and delay. Just busy.

The std::string "get" changes take some liberties that are not desirable. Specifically, assuming the string length is not the actual length and assuming ASCII encoding breaks some use cases.

Also, the utility functions are not needed because those semantics are not desired.

What was discussed was adding a single method to check length copied. Would rather see that instead.

marayl:
Thanks for getting back to me. I understand that you're busy. My main concern was that the PR would be closed due to lack of activity.

Both `std::string` and `std::str[n]len` deal with byte strings only; they make no assumptions regarding the character encoding. Some `std::string` functions assume a null-terminator, such as the `const char*` constructor, but that's about it. Take a look a `std::char_traits<char>` if you are unconvinced.

If `std::string` byte strings break some use-cases, then I'd be happy to review those use-cases along with the desired semantics if you can provide them.

It's probably worth taking a step back at this point to consider two of the main issues that I am trying to address:

1. the "put" function may result in a read overrun;
2. the "put" function does not zero pad the buffer.

Do you accept that these issues need addressing?

If my particular solution does not work for you, then I can always file separate bug reports instead, although the test-cases in my PR should still be useful for verification purposes.

tmontgomery:
Several changes made for #506  and #530 have probably invalidated this PR.

Would still rather see a single method added instead of any changes to existing methods. 

The way your functions work is not how some usages of `std::string` work to hold arbitrary data. It breaks some users current usage. I can not provide details as I don't know them. Only that I have been told they do. I suspect that things such as `std::string(const CharT*, size_t)`, which can hold arbitrary data according to constructor (4) at http://en.cppreference.com/w/cpp/string/basic_string/basic_string is probably in use.

Personally, the schema specifies a fixed length array and the app is responsible for providing it. So, read overrun and no zero padding is the apps responsibility. But I am willing to __add__ methods to aid the app as long as it doesn't break behavior that is depended upon.

marayl:
I don't think we're going to reach a consensus on this one, so I'll continue with my local work-arounds for now. I may take a fresh look in the new year when I have more time.

Issue #289, opened 2015-09-11
Generate Javadoc from description attributes


RichardWarburton:
Using description attributes to document a protocol is useful, but could be made easier to read in development code by generating Javadoc/C++ comments from those description attributes.

tmontgomery:
And also doxygen for C++. BTW, we already have this enhancement. #22

mjpt777:
The description attributes have now been added to the IR.

PR #523, opened 2017-10-22
[Java] Start of writing Javadocs

lfn3:
This is the start of work on #289/#22. I've got a couple of questions, and I thought it'd be easier to ask with the code to look at (questions are inline on the changes). Any feedback on if I'm barking up the right tree would be appreciated as well!

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -199,7 +199,7 @@ public class JavaGenerator implements CodeGenerator
 
         try (Writer out = outputManager.createOutput(className))
         {
-            out.append(generateMainHeader(ir.applicableNamespace()));
+            out.append(generateMainHeader(ir.applicableNamespace(), msgToken));
 
             generateAnnotations(indent, className, groups, out, 0, this::encoderName);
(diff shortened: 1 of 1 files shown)


sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
```diff
@@ -1366,14 +1368,36 @@ private CharSequence generateMainHeader(final String packageName)
...
 
+    private static String generateClassJavaDoc(final Token msgToken)
+    {
+        final StringBuilder sb = new StringBuilder();
+        if (msgToken.description() != null)
+        {
+            sb.append(" * ").append(msgToken.description()).append("\n");
+        }
+
+        sb.append(" * ").append("@since ").append(msgToken.version()).append("\n");
```
lfn3: Version appears to always be `sinceVersion` even when `msgToken.versionContext` indicates otherwise.
Is this right or is there some circumstance I'm missing? This is on both `BEGIN_MESSAGE` and `BEGIN_FIELD` tokens.

mjpt777: This is historical and no longer needed I believe.  Some of the early versions of SBE in the pre 1.0 phase may have used this but it is no longer.

PR #547, opened 2018-03-19
[Java] Generate put CharSequence methods for ASCII encoded fields

zyulyaev:
This PR allows to set ASCII encoded char array fields with any class implementing CharSequence, not only with Strings. It helps writing GC-free code without handling byte arrays. One for example can pass StringBuilder as an argument which can be shared between invocations.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -837,11 +837,32 @@ public class JavaGenerator implements CodeGenerator
             byteOrderStr,
             indent);
 
-        if (null == characterEncoding)
+        if (null != characterEncoding)
         {
-            return;
+            generateCharArrayEncodeMethods(
+                sb,
+                propertyName,
+                sizeOfLengthField,
+                maxLengthValue,
+                lengthType,
+                byteOrderStr,
+                characterEncoding,
+                className,
+                indent);
         }
+    }
 
+    private void generateCharArrayEncodeMethods(
+        final StringBuilder sb,
+        final String propertyName,
+        final int sizeOfLengthField,
+        final int maxLengthValue,
+        final PrimitiveType lengthType,
+        final String byteOrderStr,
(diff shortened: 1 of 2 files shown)


sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
```diff
@@ -1858,6 +1905,32 @@ private void generateCharArrayEncodeMethods(
...
+                indent + "        final int srcLength = src.length();\n" +
+                indent + "        if (srcLength > length)\n" +
+                indent + "        {\n" +
+                indent + "            throw new IndexOutOfBoundsException(" +
+                "\"CharSequence too large for copy: byte length=\" + srcLength);\n" +
+                indent + "        }\n\n" +
+                indent + "        for (int i = 0; i < srcLength; ++i)\n" +
+                indent + "        {\n" +
+                indent + "            final char charValue = src.charAt(i);\n" +
+                indent + "            final byte byteValue = charValue > 127 ? (byte)'?' : (byte)charValue;\n" +
```
tmontgomery: Is this the right thing to do with all non-ASCII encodings? Wouldn't it be better to simply copy the char value without the check?

juddgaddie:
@zyulyaev is there any reason why you did not include something similar on the Decoder side that takes an Appendable such as StringBuilder
eg
public void getMyValue(Appendable val)

zyulyaev:
@juddgaddie not really, I just added only what I needed.
Regarding Appendable to Decoder, I would rather added a method that returns a wrapper object implementing CharSequence interface, similar to how currently composite decoders work.
eg
```
private final MyValueWrapper myValueWrapper = new MyValueWrapper();

public CharSequence myValueWrapper() {
    return myValueWrapper;
}

private class MyValueWrapper implements CharSequence {
    public int length() {
        int len;
        for (len = 0; charAt(len) != 0; len++);
        return len;
    }

    public char charAt(int index) {
        final int pos = this.offset + <offset> + index;
        return (char) buffer.getByte(pos);
    }
}
```

PR #548, opened 2018-03-20
[Java] Using sinceVersion from field instead of type declaration

zyulyaev:
Generated `if` statement in decoder's methods must use `sinceVersion` from fields declaration, not type declarations.

### DIFF
--- a/sbe-tool/src/test/resources/extension-schema.xml
+++ b/sbe-tool/src/test/resources/extension-schema.xml
@@ -10,35 +10,43 @@
             <type name="templateId" primitiveType="uint16"/>
             <type name="schemaId" primitiveType="uint16"/>
             <type name="version" primitiveType="uint16"/>
-            <type name="headerLength" primitiveType="uint16"/>
-
-            <type name="extraField" primitiveType="int32" sinceVersion="5"/>
         </composite>
         <composite name="groupSizeEncoding" description="Repeating group dimensions">
             <type name="blockLength" primitiveType="uint16"/>
             <type name="numInGroup" primitiveType="uint8" semanticType="NumInGroup"/>
         </composite>
         <type name="AType" primitiveType="int32" sinceVersion="1"/>
-        <composite name="AComposite" sinceVersion="3">
+        <composite name="AComposite" sinceVersion="1">
             <type name="value" primitiveType="int32"/>
         </composite>
+        <enum name="AEnum" encodingType="uint8" sinceVersion="3">
+            <validValue name="FirstValue">0</validValue>
+            <validValue name="SecondValue">1</validValue>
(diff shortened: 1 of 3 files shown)


zyulyaev:
I also added sinceVersion static methods for composite type members.

sbe-tool/src/test/resources/extension-schema.xml
```diff
@@ -0,0 +1,44 @@
...
+                   description="Unit Test">
+    <types>
+        <composite name="messageHeader" description="Message identifiers and length of message root">
+            <type name="blockLength" primitiveType="uint16"/>
+            <type name="templateId" primitiveType="uint16"/>
+            <type name="schemaId" primitiveType="uint16"/>
+            <type name="version" primitiveType="uint16"/>
+            <type name="headerLength" primitiveType="uint16"/>
+
+            <type name="extraField" primitiveType="int32" sinceVersion="5"/>
```
tmontgomery: The [spec](https://github.com/FIXTradingCommunity/fix-simple-binary-encoding/blob/master/v1-0-STANDARD/doc/05SchemaExtensionMechanism.md) states "Message header encoding cannot change."

tmontgomery:
In general, this looks quite good. Thanks for adding the test!

tmontgomery:
Should probably remove the extension to the MessageHeader, though.

mjpt777:
Extending a message header is tricky. The message header itself does not come with its own length fields thus it will break existing code which expects message headers to be constant length. It is a part of the specification I've never been comfortable with. The fact that the schemas cannot cross reference makes this more tricky and the only practical way to do anything cross schema is to relay on fixed message headers and other types like group and var data header encodings.

zyulyaev:
Okay, I removed message header extension, but I still would like to have sinceVersion methods in composite decoder classes, if it is not a problem.
I also made similar changes to enums and bitsets.
BTW, don't you think that for enum fields it would be better to remove NULL_VAL constant instead of null reference?

mjpt777:
@zyulyaev The NULL_VAL and SBE_UNKNOWN make it easy to deal with an enum in a switch statement. Without them you could get a NullPointerException.

zyulyaev:
@mjpt777 Sorry, I meant “return”, not “remove”. And that’s exactly my point. But now generated “if” block returns null reference, not the NULL_VAL for enums.

mjpt777:
@zyulyaev Yes it would make more sense to return the NULL_VAL from the property getter if not present. Good spot.

mjpt777:
@zyulyaev I've made a change to return the NULL_VAL like you suggested.

zyulyaev:
@mjpt777 Great, thank you!

Issue #560, opened 2018-06-07
JsonPrinter and SBEIR Schema codegen issues


aksharp:
Ran into few issues while evaluating SBE 1.8.1 
1) Issues with JsonPrinter (would like to use it for logging/toolchain as it is faster than alternatives) 
- Prints invalid JSON
- Drops enum and bool fields
- Rounds up numbers
2) IR Schema file
- Enum values are missing when using sbeir schema with SBE Tool to codegen

POC code is available on github here: https://github.com/aksharp/sbe-poc

Failing test
https://github.com/aksharp/sbe-poc/blob/master/src/test/scala/sbe/JsonComparison.scala

mjpt777:
Thanks @aksharp for the report. There was indeed an issue with encoded IR which I've now fixed.

The `JsonPrinter` was introduced for debugging purposes and is incomplete. If I get time I'll look at fleshing out the remainder of the functionality.

mjpt777:
@billsegall Could you have a look at modifying the Rust codec so this test passes again and see if the C# needs any attention?

https://github.com/real-logic/simple-binary-encoding/blob/master/sbe-tool/src/test/java/uk/co/real_logic/sbe/generation/rust/RustGeneratorTest.java#L239

mjpt777:
I've now fleshed out the features to support all types.

mjpt777:
@ZackPierce Can you have a look at the Rust changes? I should be fairly simple.

ZackPierce:
Sure, I'll look into it this week.

ZackPierce:
Yep, I think I've got it sorted out. Will toss up a PR shortly.

PR #569, opened 2018-06-25
Handle IR changes to const enums for Rust; re-enable regression tests

ZackPierce:
As requested in https://github.com/real-logic/simple-binary-encoding/issues/560


mjpt777:
Thanks!

PR #570, opened 2018-06-27
Issue560 CSharp fix

billsegall:


mjpt777:
Thanks

dxwu88:
I used sbe-all-1.8.1,jar tool to generate C++ code for CME latest templates_FixBinary.xml. But the code doesn't work well. A couple of places have memory boundary issues. Does anyone have the same experience or know what the problem is? Thanks!

mjpt777:
@dxwu88 It is likely to be related to this PR. Can you test with the latest build?

dxwu88:
Sure. Thanks!

dxwu88:
For CME group message  MDIncrementalRefreshBook46 &wrapForDecode(
         char *buffer, const std::uint64_t offset, const std::uint64_t actingBlockLength,
         const std::uint64_t actingVersion, const std::uint64_t bufferLength), what should I specified for arguments actingBlockLength and bufferLength? Thanks!

mjpt777:
@dxwu88 It is not helpful to asked questions on PRs that do not help provide clarity on the PR. Best to raise questions as separate issues when unrelated.

To quickly answer the question from the above, the `actingBlockLength` is provided from decoding the message header. You can understand this by reading the SBE specification and following the examples. The `bufferLength` is provided as the length of the underlying buffer for bounds checking.

PR #575, opened 2018-07-28
Generate noChoicesSet convenience method for Sets

JerryShea:
The decoder generated for a Set does not provide a convenience method to determine if nothing has been set. If we look at sbe_samples and Car's OptionalExtras, the only way to determine if no OptionalExtras
have been set is to call 
```
    notSet = ! (sunRoof() || sportsPack() || cruiseControl());
```    
And this code is fragile and inefficient. This PR modifies SbeTool to generate a new method in OptionalExtrasDecoder
```
    public boolean noChoicesSet()
    {
        return 0 == buffer.getByte(offset);
    }
```


mjpt777:
This could be a useful feature. Would make sense to call it something more like `isEmpty()` and it also needs to include taking care of the null value when optional.

mjpt777:
Please run the full build to ensure it passes.

JerryShea:
Woops - fixed now.

PR #578, opened 2018-07-31
golang: Use pointer receiver where appropriate

lkramer:
- In order to avoid copying the whole structure functions should use pointer receiver when possible.
- One exception is for enums, where it makes more sense to keep the copy receivers.


mjpt777:
Maybe @billsegall can comment on this?

billsegall:
I'm happy with this. It's 100% compatible and benchmarks faster on Encode. Good stuff.

PR #584, opened 2018-09-13
Add Travis ci integration

rigtorp:
This will build and run the tests for the C++ implementation.


mjpt777:
What else is required to enable this on Travis? I'm getting a "this is not an active repository" error.

mjpt777:
I notice that Travis requires read and write access to my repos. Why the write access?

PR #589, opened 2018-10-01
In/outgoing Span<byte> support, linux net45 build

### LINKED ISSUE #587: [C#] Support Span<T> instead of / on top of T[] for array getters/setters
Hi,
I'm opening this issue to start a minimal discussion before I go ahead and attempt to implement a change to the code generation that sbe does for C#.

C# has support for [`Span<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=netcore-2.1) which can be used to encapsulate various types of array of T where those can be:
* Unmanaged memory
* stack allocated memory
* Existing managed / GC memory

If we take the generated code for Car from: https://github.com/real-logic/simple-binary-encoding/blob/master/csharp/.nuget/examples/baseline/Car.cs

Here's a short relevant excerpt:

https://github.com/real-logic/simple-binary-encoding/blob/8ec67b1c28ce70b845ad43839f309dc7bf7b5473/csharp/.nuget/examples/baseline/Car.cs#L964-L986

We can clearly see that SBE currently generates `Get`/`Set` method pairs that can take as input/output a managed array of the underlying type (byte in case of a string).

I would like to either replace or add additional Get/Set methods that accept a `Span<byte>` in this case, or a `Span<T>` in the general case.

Adding such Getters/Setters would allow users to copy strings / arrays out of the message into stack allocated memory, for example, without requireing to have a managed buffer as it currently required, for example in the example usage for the same sample:
https://github.com/real-logic/simple-binary-encoding/blob/8ec67b1c28ce70b845ad43839f309dc7bf7b5473/csharp/.nuget/examples/CarExample.cs#L184

https://github.com/real-logic/simple-binary-encoding/blob/8ec67b1c28ce70b845ad43839f309dc7bf7b5473/csharp/.nuget/examples/CarExample.cs#L259

I think this could be a nice first step in improving SBE's friendliness for high-perf / zero allocation users.

As a second step I ALSO think it would be great if additional `GetXXX()` methods would be introduced returning a `Span<T>` pointing to the internal memory, but I would first want to get this done before opening a separate issue on that matter and implementing that.

Any thoughts?

damageboy:
- Closes #587
- Add netfx.props and tweak sbxee-dll csproj to allow building sbe.dll
  for net45 on Linux using dotnet sdk following:
  https://andrewlock.net/building-net-framework-asp-net-core-apps-on-linux-using-mono-and-the-net-cli/
- Remove packages.config files from projects already converted to SDK style project since
  those files serve no purpose other than confusing potential contributors...
- Intoroduce minimal Span<byte> GetBytes/SetBytes implementation for copying
  to/from Span<byte> to sbe.dll itself.
- Change code generation for array types to use Span<byte> internally keeping the
  original functionality intact with no (hopefully!) visible user facing changes except
  for the addtitional Get/Set methods accepting Span<byte> being added to the mix.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/csharp/CSharpGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/csharp/CSharpGenerator.java
@@ -358,16 +358,21 @@ public class CSharpGenerator implements CodeGenerator
                     sizeOfLengthField));
 
                 sb.append(String.format("\n" +
-                    indent + "public int Get%1$s(byte[] dst, int dstOffset, int length)\n" +
+                    indent + "public int Get%1$s(byte[] dst, int dstOffset, int length) =>\n" +
+                    indent + INDENT + "Get%1$s(new Span<byte>(dst, dstOffset, length));\n",
+                    propertyName));
+
+                sb.append(String.format("\n" +
+                    indent + "public int Get%1$s(Span<byte> dst)\n" +
                     indent + "{\n" +
                     "%2$s" +
                     indent + INDENT + "const int sizeOfLengthField = %3$d;\n" +
                     indent + INDENT + "int limit = _parentMessage.Limit;\n" +
                     indent + INDENT + "_buffer.CheckLimit(limit + sizeOfLengthField);\n" +
                     indent + INDENT + "int dataLength = (int)_buffer.%4$sGet%5$s(limit);\n" +
-                    indent + INDENT + "int bytesCopied = Math.Min(length, dataLength);\n" +
+                    indent + INDENT + "int bytesCopied = Math.Min(dst.Length, dataLength);\n" +
                     indent + INDENT + "_parentMessage.Limit = limit + sizeOfLengthField + dataLength;\n" +
-                    indent + INDENT + "_buffer.GetBytes(limit + sizeOfLengthField, dst, dstOffset, bytesCopied);\n\n" +
+                    indent + INDENT + "_buffer.GetBytes(limit + sizeOfLengthField, dst.Slice(0, bytesCopied));\n\n" +
                     indent + INDENT + "return bytesCopied;\n" +
                     indent + "}\n",
                     propertyName,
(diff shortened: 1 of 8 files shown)


damageboy:
It's worthwhile to mention it might *appear* that I've removed a couple of input validation checks in the generated code, but this is not really so. Those checks happen anyway inside the constructor of `Span<T>` and will throw essentially the same exception from inside the `Span<T>` ctor.

damageboy:
Any comments about this?

billsegall:
My only question would be whether there was test cases for those exceptions?

Bill

ZachBray:
Thanks for the PR. Looks good to me, in general. :-)

My understanding is that only the code generation happens inside the Gradle script not the building/testing of the C# solution.

I have a few questions. I'm not sure who is best placed to answer them (maybe @billsegall ?).

1. Could the introduction of the `System.Memory` dependency affect anyone? I.e., could users have been using `.NET Framework < 4.5` to build the SBE generated code prior to this change? I guess the lower limit is currently whatever [the sbe-tool package](https://www.nuget.org/packages/sbe-tool/0.1.8.1-beta-1) targets. The NuGet site doesn't make this obvious for this particular package. 

2. If users are affected, are we still happy to make the change? Or should we introduce a switch into the code generation that enables Span support? I think Olivier suggested this in #587.

3. Do we use any features in `netstandard2.0` that aren't in `netstandard1.1`? I ask as it looks like `netstandard1.1` is supported by [the System.Memory package](https://www.nuget.org/packages/System.Memory/) and according to [the .NET Standard compatibility matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support) this is supported by .NET Framework v4.5. Would this mean we could simplify the build somewhat?

ZachBray:
Also...

4. Do we keep release notes somewhere?

billsegall:
We're currently targeting the nuget packages at netstandard2.0 and have
dropped 1.1. No-one complained but that doesn't mean they won't. In general
I'm happy to maintain a reasonable amount of backward compatibility but MS
made this very difficult pre 2.0.

Similarly 4.5 would seem to me to be a reasonable minimum at this point but
I'd be happy to address this if anyone really needed it. In the absence of
a demonstrated need I'd leave things at that,

Bill.





Bill.

billsegall:
Release notes are a good idea. How do people feel about an additional
section in ~csharp/README.md?

Bill.

damageboy:
I was thinking about (per @mjpt777) to update the relevant wiki pages re. the new dependency post merge.

I didn't think about the README.md inside the csharp folder, and see no reason not to amend this commit to reflect that dependency as well..

mjpt777:
The bulk of the notes can be in the csharp/README.md which can be referenced from the wiki.

ZachBray:
@billsegall 

> I'm happy to maintain a reasonable amount of backward compatibility but MS
made this very difficult pre 2.0

I didn't realise that. Fair enough! :-)

> We're currently targeting the nuget packages at netstandard2.0

Assuming the consumers are not building from source (is this a fair assumption?), this means we support `.NET Framework >= v4.6.1` according to the .NET Standard compatibility table. In which case, please can we drop the multiple targets from here?

https://github.com/real-logic/simple-binary-encoding/blob/master/csharp/sbe-dll/sbe-dll.csproj#L4

I.e., only target `netstandard2.0` for the class library. I believe this will make it easier to build on unix without mono if we switch the tests to use dotnet core. We could then get the CSharp tests running as part of [the Travis build matrix](https://github.com/real-logic/simple-binary-encoding/blob/master/.travis.yml).

ZachBray:
@damageboy apologies for polluting your PR with these comments! They probably belong somewhere else.

damageboy:
@ZachBray  As I'm now knee deep in figuring out how the unit tests work, it's actually good timing I think.

To be quite honest, I think I should simply run the tests as is manually on windows and give a scout's honor promise the tests are green.

Other than that, I think the tests require their own major overhaul in a separate PR,
I've opened a different issue for this: https://github.com/real-logic/simple-binary-encoding/issues/592

mjpt777:
Can @JPWatson, @odeheurles, or @billsegall let me know if this should be merged or not?

billsegall:
I've completed a manual build/test with the patch and it's working for me
on Windows building with VS2017. I'll probably freshen the nuget package
following the release of 1.9.1.  I didn't add any validation tests using
the the new code.

Bill.

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/csharp/CSharpGenerator.java
```diff
@@ -756,44 +766,30 @@ private CharSequence generateArrayProperty(
...
-            propName,
-            fieldLength));
+            propName, fieldLength));
 
         sb.append(String.format("\n" +
             indent + "public %1$s Get%2$s(int index)\n" +
             indent + "{\n" +
-            indent + INDENT + "if (index < 0 || index >= %3$d)\n" +
-            indent + INDENT + "{\n" +
+            indent + INDENT + "if (index < 0 || index >= %3$d) {\n" +
```
mjpt777: Why change the generated format here? Will this not now be inconsistent?

Issue #615, opened 2018-11-08
MaxValue for uint32 type validated against int32.max


kdkavanagh:
Checking to see if this validation implementation is intentional.  My expectation was that a uint32 max value would be checked against uint32.max, not int32.max

From CompositeType.java:
```java
            final long maxInt = INT32.maxValue().longValue();
            if (primitiveType == UINT32 && longValue > maxInt)
            {
                XmlSchemaParser.handleError(node, String.format(
                    "maxValue greater than allowed for type: maxValue=%d allowed=%d", longValue, maxInt));
            }
```

kdkavanagh:
Guess based on the related generated code for dimension en/decoders that there is some requirement that both blockLen and numInGrp are always <= int32. 

Related to #609

mjpt777:
This is correct. The spec recommends `UINT8` and `UINT16` for the types. We allow a bit of flexibility in going larger but given the limitations of the Java language we limit to max signed int.

kdkavanagh:
SBE also enforces unsigned types for numInGrp.  Should this restriction be relaxed to allow 2^31 (int32) objects in a repeating group?  While semantically a signed count doesn't make a ton of sense, there is a big difference between 2^16 and 2^31

mjpt777:
That is a question for the spec more so than this implementation.

Issue #609, opened 2018-11-01
Compilation errors when specifying large Repeating Groups


RichardWarburton:
If you want to encode repeating groups larger than a `uint16` then you need to specify an increase in the size of the block length property:

```
<composite name="groupSizeEncoding" description="Repeating group dimensions" >
    <type name="blockLength" primitiveType="uint32"/>
    <type name="numInGroup" primitiveType="uint16" semanticType="NumInGroup"/>
</composite>
```

But if you do this then SBE will generate Java code with a compilation error in:

```
RedactedDecoder.java:121: error: incompatible types: possible lossy conversion from long to int
            blockLength = dimensions.blockLength();
```

I can understand if you don't wish to support block lengths larger than a `uint16` but I think doing this with SBE validation rather than generated invalid Java code would be a better approach.

PR #617, opened 2018-11-08
Warn instead of error on signed numInGrp, check for minValue < 0

kdkavanagh:
Relates to comment on #615.  Allow signed numInGrp and validate that the minValue is > 0


mjpt777:
The type must be unsigned according to the specification.

kdkavanagh:
Is this materially different than #609?  I don't see the harm in flagging it with a warning but still allowing it to overcome java's unsigned type limitation specifically - All code properly implementing the spec (i.e minValue) should handle this correctly.

kdkavanagh:
Thanks!

mjpt777:
You made a good point :-)

PR #629, opened 2019-01-30
Header validation warning on non-uint16 types instead of error

kdkavanagh:
SBE generator currently errors out if schema/template/version fields are not 16 bits.  This PR changes that error to a warning to match the existing validation for blockLength

Similar to #609 and #617


tmontgomery:
@mjpt777 I think we should relax this. Maybe add a `strict` mode where this is an error. But a warning instead for normal usage?

PR #631, opened 2019-01-30
[c++] Improved field generation

ksergey:
Added std::string_view getter/setter for string-like fields
Added non-const raw access for array like fields

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -1183,7 +1183,17 @@ public class CppGenerator implements CodeGenerator
             token.arrayLength()));
 
         sb.append(String.format("\n" +
-            indent + "    const char *%1$s() const\n" +
+            indent + "    const char *%1$s() const SBE_NOEXCEPT\n" +
+            indent + "    {\n" +
+            "%2$s" +
+            indent + "        return (m_buffer + m_offset + %3$d);\n" +
+            indent + "    }\n",
+            propertyName,
+            generateTypeFieldNotPresentCondition(token.version(), indent),
+            offset));
+
+        sb.append(String.format("\n" +
+            indent + "    char *%1$s() SBE_NOEXCEPT\n" +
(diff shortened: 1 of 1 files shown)


mjpt777:
I have merged but this has some issues. What if the string or string view is shorter than the array? The Java codec deals with this by zero padding the remainder and doing a min on the copy for length of view and length of array.

ksergey:
>>What if the string or string view is shorter than the array?
It's UB. Actually I added `std::string_view` in the same way as `std::string` was added. So, user should solve this case by themself.

ksergey:
I think we could add deactivatable via macros check `value.length() == fieldLength()` inside `putField(...)` methods.

mjpt777:
Identifying this as UB feels wrong and error prone. The existing string approach was also wrong in this.

tmontgomery:
`std::string_view` is C++17. I don't think we want to go there just yet.

mjpt777:
@tmontgomery the code is guarded for C++ 17 so only compiled when available.

tmontgomery:
And, yes, if the string is short, it should null pad.

tmontgomery:
@mjpt777 yes, I understand that. I would, though, say that to avoid noise and the fact that there is very little reason to use it yet that there is no compelling reason to do it at this time.

ksergey:
@tmontgomery why short string should be null padded? Is string getter should be aware of this (i.e. return string with length less than field length)?

tmontgomery:
It probably should be encoding specific what to pad with. And it should be compatible with what Java does right now.

The getter is always going to be a specific length. It just may have encoding specific nulls, etc. The setter is the one to determine what to do with the string/string_view is too short.

mjpt777:
The spec example suggests null padding.

https://github.com/FIXTradingCommunity/fix-simple-binary-encoding/blob/master/v1-0-STANDARD/doc/02FieldEncoding.md#examples-of-fixed-length-character-arrays

tmontgomery:
I've added length checks and handling via 825e0c48

tmontgomery:
checks for `const char *` versions are not needed. Those are the sharp knives. Know how to use them. `std:string` and `std::string_view` are for those who need their hands held while they cross the street.

PR #633, opened 2019-01-30
Added string length check for put-string methods.

ksergey:


ksergey:
Also need to do something with `putField(const char*)`. There is no way to check string length without `strnlen`.

tmontgomery:
This is not the correct way to do this so that it is consistent with what Java does. Appropriate checks have been added. Closing this.

PR #657, opened 2019-03-18
Simplify generated C code

denizevrenci:


mjpt777:
Have you noticed how many warnings there are in GCC for the C tests?

https://travis-ci.org/real-logic/simple-binary-encoding/jobs/510877501

denizevrenci:
Thanks for pointing that out. It might make sense to treat warnings as errors to catch them during CI.

I've checked the generated code and it seems to me that this is a false positive as long as `SBE_NO_BOUNDS_CHECK` is undefined. The group in question is created uninitialized but the call to `wrap_for_encode` tries to set values for dimensions before actually using them. And if the method fails due to short buffer length, it stops execution.

mjpt777:
What should be done to address it?

PR #659, opened 2019-03-26
Fix warnings in tests for C generator

denizevrenci:


mjpt777:
The build is failing with warnings.

denizevrenci:
For some reason, I do not get these warnings in my GCC when building locally. So, the only way to test for me is to send it to the CI.

mjpt777:
Thanks

Issue #665, opened 2019-04-06
Python generator implementation


feribg:
I just finished work on a pure python stub generator. Right now  it's somewhat crude and quality of generated code can be improved upon but it's more or less fully functional. Wondering if you're interested in a PR. It's based on the Csharp generator, so it follows the single stub with multiple methods `wrap_encode`, `wrap_decode` instead of  2 classes as per the reference Java impl. It also has a couple of neat things like  full type hints and usage of https://docs.python.org/3/c-api/buffer.html#bufferobjects , however they're somewhat muted  because python has to cast most objects to int anyways, a Cython impl should be trivial based on the current code though.

mjpt777:
We would be happy to discuss a PR. However we have seen partial implementations in the past go nowhere so we are only really interested in languages that will be completed and supported longer term.

We have learned over time that separating the encoding and decoding into separate classes is more appealing and results in safer code.

feribg:
Let me know if I can  clarify or tidy it up in some ways

samratbatth:
Any eta on python support?
Is this supported https://github.com/tfgm/sbedecoder ?

feribg:
@samratbatth To be honest I haven't had time to put towards it, but if you want feel free to clone it and clean it up, it's mostly code cleanup not functional issues. 

Although if im to write it up again, I would likely just wrap the C++ codecs with pybind or something like that. The `struct` library is slow so if we can reuse the buffers and cast to `numpy` types directly that will likely give you C like speed in Python.

sahilshetty13:
Any suggestions on implementing SBE decoder in Python for CME MDP 3.0 market data?

PR #667, opened 2019-04-10
[Java] Python generator implementation #665

feribg:
This is a very early implementation of a python stub generator. It's derived from the C# Java generator, so it follows most of the code and conventions adopted there.

The `runPythonTests` gradle task will build the python stubs in the project_root/python/tests/gen and run the unit tests.  So far only car_example_schema.xml is implemented, but happy to port any other tests that focus on specific edge cases (Maybe a FIX sample).

Built and tested with python 3.7, but in practice any version that supports https://docs.python.org/3/c-api/buffer.html should work, so 2.7 should be fine too, although I didn't try it yet (also I much rather not offer support for it generally).

If you think it's not worth it adding and maintaining that code in the mainline I'm happy to fork off only the IR and Java generator as well and providing it as an unsupported 3rd party generator, depending on the discussion.


mjpt777:
This PR fails the build. As a minimum it should pass all tests.

mjpt777:
Build is still failing with lots of code style issues.

feribg:
I installed the checkstyle idea plugin imported the style and reformatted  everything, but i get tons of errors on indent and line length. Is that an idea issue not respecting fully the checkstyle config? To the issues with indents in many cases. I attached an example of what the IDE does in the second screenshot when reformat is applied with the checkstyle, but then checkstyle reports that as incorrect.
I think the rest that was wrong in terms of temp folders, etc, should be resolved.

Screen Shot 2019-04-10 at 2 37 40 PM
Screen Shot 2019-04-10 at 2 37 31 PM

mjpt777:
IDEA does not properly support strict indenting.

feribg:
Got it, okay it'll take me a while to clean them up manually then. I'll push a fix once the checkstyle passes. Meanwhile the rest should work (build, test, etc), without the style.

feribg:
I think everything is more or less cleaned up, could use more unit tests, but will likely postpone for after initial feedback.

mjpt777:
Looks reasonable. You probably want to clean up the copyright and C# references in Javadoc.

The major point for me is that is follows the combined encoder and decoder format of the C# generator and not the separation in the Java generator. We've seen this is a much better approach as the separate decoder can deal with read only buffers.

feribg:
I  tend to agree with that  point. I  already had  the issue of having to  use mutable buffer types. I'll have a look into how much effort it is to split them. If it  turns out to be too big of a rewrite are you OK merging as is contingent on cleanup. 

Are there any specific edge cases you can think of in general for which it would be  a good  idea to  port  the tests?

mjpt777:
Have a read back through the issues to get a feel for what to watch out for.

feribg:
Added some new test cases, additional fixes, the cleanups you suggested and some fixes post rebase.  Seems line one commit from Apr 9, diverged the` example-schema.xml` form the `sbe-samples` and `sbe-tools/tests`.

feribg:
Just wondering if you had a chance to take a peak at merging that?

mjpt777:
As mentioned above we want to move away from the use of a single codec for encode and decode rather than propagate this pattern. Can you look at doing something similar to the Java codec?

feribg:
Yep I will try and put some cycles towards that,  will give you an update when i have a cleaner prototype.

feribg:
Sorry about the big delay here, got side tracked. This is fully reworked to have enc + dec pattern like the Java version. I used it  as the base blueprint. Waiting for travis to complete, let me know how it looks like.

feribg:
@mjpt777 Wondering if you're still considering merging this upstream or keep it as a fork for Py support? Thx!

mjpt777:
Sorry but I had to revert this. As I worked through the code refactoring it I noticed far to many warnings and have concerns about the completeness. Please resubmit with less warnings and try not to have unused methods.

PR #678, opened 2019-05-06
Configure build scan plugin

aalmiray:


mjpt777:
Rather than include the question on build it is nicer to allow the user to pass the `--scan` option like the CI does.

aalmiray:
If you run a build with `--scan` without agreeing to the terms of service then you'll get asked every single time. The current setup asks you at least one time to agree to the terms. After that you can run a local build with `--scan` and no question should pop up.

You can simplify the build by accepting the terms on behalf of anyone that makes a build but some may object to that.

mjpt777:
It would be nice if the build scan question is only asked if a scan is being run. When you don't care about scans then the question is annoying.

mjpt777:
To be clear the scan statement would be best if not output unless a scan is run.

aalmiray:
Let me think about that, it may be the case that agreeing on the terms of service too late in the process won't have the desired effect.

mjpt777:
It does not ask if no scan is run but it makes for noisy output. Feels like the statement about setting up acceptance should only be output if a scan is run and no acceptance is in place.

mjpt777:
BTW I had a look at a few scans and it is useful for seeing where time is spent. Thanks.

mjpt777:
Is there a means of telling if a scan has been requested so that the statement can be conditionally printed?

aalmiray:
I think it's possible to determine if a scan was requested, i.e, the `--scan` command flag was given, but not with the current version of the `com.gradle.build-scan` plugin. There's some extra code required for that. This feature could be added to https://aalmiray.github.io/kordamp-gradle-plugins/#_org_kordamp_gradle_plugin which simplifies configuration of the base `com.gradle.build-scan` plugin.

PR #681, opened 2019-05-09
[C] Remove the dimension member from the codec classes

denizevrenci:


mjpt777:
The build is failing on this one.

denizevrenci:
It looks like there are more of the "maybe uninitialized" false positives. I will push a fix.

PR #685, opened 2019-05-31
enhancements to Rust generator

ian-p-cooke:
while experimenting with the Rust generator I found a few things that didn't work as expected:

* cargo_exits() in build.gradle wasn't finding cargo
    * the 'args' argument should be an array of arguments
    * was using `-v` (verbose) instead of `-V` (version).  `-V` is valid by itself while `-v` is not.
    * the output went to the screen 
* warnings during `gradlew.bat test`
    * caused by a type identifier that was invalid for C#

I cleaned those up and added the ability for rust to decode a header and then pick a template by `template_id`.  This required making some things pub which were private before but I don't think that's an issue.  The code for using this new method looks like this:
```Rust
let (h, buf) = start_decoding_message_header(&buffer)?;
match h.template_id {
    CarMessageHeader::TEMPLATE_ID => {
        let dec_fields = CarFieldsDecoder::wrap(buf);
        //...
    },
    _ => panic!("unknown template_id");
}
```

If keeping the `FieldsDecoder` and `ScratchDecoderData` private is really important then maybe some helper that skips over the bytes of decoded header would be sufficient.  The implementation here was just a 'get it to work for me' attempt and conversation-starter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@
 # Build tools
 build-local.properties
 deps
-/bin
+*/bin
 build
 .gradle
 target
@@ -23,6 +23,7 @@ GPATH
 prop
 out
 .vs/
+.vscode
 
 # cmake
 
--- a/build.gradle
+++ b/build.gradle
@@ -569,7 +569,8 @@ def cargo_exists() {
     try {
         def result = project.exec {
             executable = 'cargo'
-            args = '-v'
+            args = [ '-V' ]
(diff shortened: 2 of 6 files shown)


mjpt777:
@ian-p-cooke if you address the conflict then I'm happy to merge.

ian-p-cooke:
> @ian-p-cooke if you address the conflict then I'm happy to merge.

ah, thanks, I didn't see the conflict.  reverted my commit that did the same thing as yours.

PR #689, opened 2019-06-10
Add new compilers

rigtorp:
This will cover everything except Redhat 7 I think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: xenial
+
 sudo: false
 
 matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,21 +9,31 @@ matrix:
       language: java
       jdk: openjdk8
       os: linux
-      dist: xenial
```
mjpt777: Why was this changed to default? The default for Travis is 14.04 when most people are on 16.04 or later.

tmontgomery: Correct me if I am wrong.... Xenial is 16.04 LTS. Isn't it the latest LTS? Bionic is 18.04 LTS. Not sure it's image in Travis is that stable yet, though.

mjpt777: @tmontgomery  You are correct.  However Travis default is Trusty. https://docs.travis-ci.com/user/reference/overview/

tmontgomery: Trusty being 5 years old.... shouldn't it be compiling against a newer version? I'm just making the argument as I see it. I don't know if the PR author had any other reason or reasons. Maybe @rigtorp can weigh in.

mjpt777: I changed it to be Xenial to be more recent and now it is reverted.

tmontgomery: Isn't xenial now the default at the top? Making it unnecessary to specify it on each one?

mjpt777: Missed that. Sorry.

PR #699, opened 2019-07-18
Reorg C# code for 2019

damageboy:
- Unit tests actually pass now
- runtest.sh will generate a test file from the java sample and use that
  for interop tests rather than hardcoding the byte stream
- travis builds the project, runs tests and generates a nuget (that goes
  no-where for now)
- nuget generated from .csproj, no more clunky release script or manual
  .nuspec
- samples split into folders


damageboy:
I think this PR is now ready to merge...

@billsegall @JPWatson I would appreciate if you guys could release a new nupkg onto nuget.org after running the `pack.sh` script and making sure I didn't mess up anything...

mjpt777:
@damageboy  What is the `ExampleUsingGeneratedStub.sbe` in the project root directory?

damageboy:
Well, the previous iteration had an interop unit-test with java, that had a literal byte-array directly embedded in the code...

That array was in effect the result of one of the java samples being run in such a way it would output a file, then that file was translated to the byte array...

I really didn't like that, and it's likely to break whenever something changes anyway since the code-generation generates a different encoder/decoder, but the sample is not kept up to date...

So I added [this](https://github.com/real-logic/simple-binary-encoding/blob/04a471a6a045cdbddef3d3f715644c17f9e06110/csharp/runtests.sh#L4) to the `runtest.sh` to generate that file as part running the tests in travis...

Users who run this tests not as part of travis will simply automatically skip the test as it is marked as inconclusive if the `.sbe` file does not exist.

Also, the `.gitignore` has [this](https://github.com/real-logic/simple-binary-encoding/blob/04a471a6a045cdbddef3d3f715644c17f9e06110/.gitignore#L101) ignore rule that makes sue the `.sbe` file is not commited...

Is that OK?

mjpt777:
If needed then best to store in a csharp related sub-directory.

mjpt777:
BTW I'm referring to this file. https://github.com/real-logic/simple-binary-encoding/blob/master/ExampleUsingGeneratedStub.sbe

damageboy:
@mjpt777 Sorry, this can be deleted.
I think it's an earlier attempt that I committed by accident.

The code DOES store it in the csharp specific directory where it is ignore with `.gitignore`.

My apologies to committing that erroneously.

mjpt777:
No worries. I've removed it.

PR #707, opened 2019-08-16
Revert the package name to sbe-tool for release 1.13.1

billsegall:


mjpt777:
Is there a change we should make to the release process to update the version? Can the version be read from the [version.txt](https://github.com/real-logic/simple-binary-encoding/blob/master/version.txt) file?

PR #731, opened 2019-10-24
[Java] Avoid memory allocation at varDataDisplay

ariksher-nex:
varDataDisplay - use method `getVarFieldName(Appendable)` - no memory allocation. unlike calling method `varFieldName` where it allocates memory (new byte[] and new String).

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -3274,7 +3274,9 @@ public class JavaGenerator implements CodeGenerator
                     append(sb, indent, "builder.append('\\'');");
                     append(sb, indent, formatGetterName(varDataToken.name()) + "(builder);");
                     append(sb, indent, "builder.append('\\'');");
-                } else {
+                }
+                else
+                {
                     append(sb, indent, "builder.append('\\'').append(" + varDataName + "()).append('\\'');");
                 }
             }
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -904,7 +904,7 @@ public class JavaGenerator implements CodeGenerator
                     generateGet(lengthType, "limit", byteOrderStr),
                     byteOrderStr);
             }
-            if (characterEncoding.contains("UTF8"))
+            if (characterEncoding.contains("UTF"))
             {
                 new Formatter(sb).format("\n" +
                     indent + "    public int get%1$s(final Appendable appendable)\n" +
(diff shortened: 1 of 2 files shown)


mjpt777:
This generates code which does not compile.

ariksher-nex:
> This generates code which does not compile.

yep, fixing it. apperantly code differs for UTF and ASCII.

mjpt777:
The UTF-8 implementation allocates too. I'm not in favour of supporting this as it is too brittle.

ariksher-nex:
> The UTF-8 implementation allocates too. I'm not in favour of supporting this as it is too brittle.

I'm using at my project ASCII chars and I'm willing to use the method **appendTo** that sits on the generated xxxDecoder class(es), which is very useful for logging incoming messages.

mjpt777:
I've pushed a more obvious changes which saves on allocation.

ariksher-nex:
> I've pushed a more obvious changes which saves on allocation.

seen your commit and it paratially solves the problem see my comment there
https://github.com/real-logic/simple-binary-encoding/commit/391527ac04df032671167b12d923b29dad30c5c0#r35648950

mjpt777:
A better approach for you to take is to use the append version if and only iff the encoding is ASCII otherwise output as is at present.

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
```diff
@@ -904,6 +904,28 @@ private void generateDataDecodeMethods(
                     generateGet(lengthType, "limit", byteOrderStr),
                     byteOrderStr);
             }
+            if (characterEncoding.contains("UTF"))
```
mjpt777: Why is this method still present?

ariksher-nex: removed it.

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
```diff
@@ -3269,7 +3269,14 @@ private void appendDecoderDisplay(
             }
             else
             {
-                append(sb, indent, "builder.append('\\'').append(" + varDataName + "()).append('\\'');");
+                if (characterEncoding.contains("ASCII"))
+                {
+                    append(sb, indent, "builder.append('\\'');");
+                    append(sb, indent, formatGetterName(varDataToken.name()) + "(builder);");
+                    append(sb, indent, "builder.append('\\'');");
+                } else {
```
mjpt777: This will fail formatting.

ariksher-nex: fixed the checkstyle thing :-)

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
```diff
@@ -3269,7 +3269,14 @@ private void appendDecoderDisplay(
             }
             else
             {
-                append(sb, indent, "builder.append('\\'').append(" + varDataName + "()).append('\\'');");
+                if (characterEncoding.contains("ASCII"))
+                {
+                    append(sb, indent, "builder.append('\\'');");
```
ariksher-nex: @mjpt777 do the single quotes around the string required?
care if i take those out? (both in ASCII and regular implementation)

mjpt777: Best to leave them.

PR #738, opened 2019-11-29
[C++] Removed property access in ostream printer

duxi90:
* Removed generated access of a property when character encoding
is not set
* This fixes unused return value warning


mjpt777:
This fixes a warning but breaks the functionality. The field needs to be read to be skipped over. I'm going to push a better implementation.

PR #749, opened 2020-02-10
[Java] Migrate to `maven-publish` plugin and cleanup artifacts being published

vyazelenko:
The following changes were done to the generated artifacts:
- `sbe-tool`:
  * exclude Golang sources from all the jars
  * include C++ oft headers
- `sbe-all`:
  * exclude Golang sources from all the jars
  * include C++ oft headers
- `sbe-samples`:
  * include generated sources into the jars
  * include proper MANIFEST metadata

Other changes include:
- Cleanup build to use helper functions
- Configure HTTP timeouts and retries via Gradle system properties

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ plugins {
     id "com.github.ben-manes.versions" version "0.27.0"
 }
 
-defaultTasks 'clean', 'build', 'shadowJar', 'install'
+defaultTasks 'clean', 'build'
 
 def checkstyleVersion = '8.28'
 def hamcrestVersion = '2.2'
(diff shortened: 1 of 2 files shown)


billsegall:
Excluding golang/templates will break golang.

Bill.

vyazelenko:
@billsegall we want exclude everything non-java from the published artifacts while retaining ability to generate Go code from the project sources. Do you see any problems with this approach?

vyazelenko:
I've fixed Go code generation with https://github.com/real-logic/simple-binary-encoding/commit/027fe21e865cc135fbda5239cec7d2f47f955610

Issue #291, opened 2015-09-11
Add a method to calculate the length of an SBE message


RichardWarburton:
If you're using an API which requires you to allocate or claim a block of memory in order to write the SBE message into it then you need to know the length of the encoded message.

This can be calculated by BLOCK_LENGTH + sum(length of each variable data field + size of length field of each variable data field).

However this couples users of the API to the details of the encoding and requires them to know that the size of length fields aren't part of the BLOCK_LENGTH. Having a static `requiredLength()` method would make the API easier to use.

donmendelson:
The proposed method should also account for repeating groups. One query is: what is the required memory for a message with X repeating group entries. The inverse is: how many repeating group entries can be held by a buffer or datagram of length Y. This would be useful to help maximize bandwidth utilization.

mjpt777:
To make this work we will need a signature that takes the number of repeating groups and a collection of all var data items. This could get quite complicated with var data in repeating groups, or repeating groups within repeating groups.

In the case of var data containing strings that needs to be encoded to a variable width encoding like UTF-8 then we would need to do the encoding before determining the length, or conservatively estimate.

Is there a possible minimum set that would be useful?

PR #761, opened 2020-03-04
[C++] Methods to calculate the length of an SBE message

agmt:
Implements: https://github.com/real-logic/simple-binary-encoding/issues/291


mjpt777:
Will look in more detail in at a later date. Got a bit too much on at the minute.

One thing on naming. Something that will calculate the total so be named to reflect that it is not a simple accessor. Something like `computeEncodedLength`.

agmt:
@mjpt777 noticed. New function names are: `static computeLength(...)` (to predict message length before encoding) and `decodeLength() const` (to find length in the beginning of decoding).

mjpt777:
I see you were fixing some warnings in addition to this. I've picked those up independently.

agmt:
I've rebased on master

mjpt777:
Be good to get @tmontgomery view on this.

PR #770, opened 2020-03-26
Refinement the C generator

lygstate:
Add sbec.h and remove redundant types
Add _as_string_view fro all string
typedef struct x x;
Add retrieve enum value directly function
memcpy need size_t
Place diagnostic to the file begin and end
Move dimensions to the beginning


tmontgomery:
Adding a shared header is not something we want to do at this time as mentioned previously. At the very least that should be separated from the other refinements.

mjpt777:
It is clear from CI that this PR has not even had the build run locally. I'm tempted to close it.

mjpt777:
I'm closing this as the build fails. It can be resubmitted if feedback is taken and the build passes.

PR #772, opened 2020-03-29
[C] Add unsafe function for enum property get, and merge three enum r…

lygstate:
…elate function into single String.format


mjpt777:
What is the purpose of the PR? Please provide more details. We will not merge PRs without a clear purpose. New features should also come with tests.

mjpt777:
Closing due to no response to question.

PR #774, opened 2020-03-29
The return type of generateEnumFieldNotPresentCondition are enum, but…

lygstate:
… the return type of function %7$s_%2$s are bool, they are in-consistence.

Remove generateEnumFieldNotPresentCondition first, maybe have a better resolution.


mjpt777:
A correct fix has been applied.

PR #777, opened 2020-05-04
[C#] Add Offsets and WrapForEncodeAndApplyHeader

szehetner:
This adds two new members to the generated C# messages. Both of these are already available in the generated Java classes, but were missing in the C# version.

**`public const int <FieldMember>Offset`**
This is useful if you need to read/write a member with some other method than the generated accessors, e.g. read a fixed-length byte array as a `ReadOnlySpan<byte>`. Currently these offsets need to be hardcoded.

**`public void WrapForEncodeAndApplyHeader(DirectBuffer buffer, int offset, MessageHeader headerEncoder)`**
Generates the repetitive code to fill the header. Same as "wrapAndApplyHeader" in Java.


vyazelenko:
@szehetner The build fails with the Checkstyle error, i.e.:
```
sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/csharp/CSharpGenerator.java:1038: Line is longer than 120 characters (found 137). [LineLength]
```

szehetner:
@vyazelenko Thanks, forgot to run the full gradle build locally.

Issue #728, opened 2019-10-04
[C#] Generated SchemaVersion field in wrong place


rca22:
When generating C# stubs, the `version `attribute of the `sbe:messageSchema` XML element becomes a property for each of the generated classes. This is logically incorrect, as to parse a message you would typically do the below (as per your own documentation)
```
// first we decode the header (in a real world scenario you would need the header to decide which SBE decoder you are going to use
bufferOffset = 0;
// position the MessageHeader object at the beginning of the array
MessageHeader.Wrap(directBuffer, bufferOffset, Car.SchemaVersion);

// Extract info from the header which you would use to lookup the
// applicable flyweight to decode this type of message based on templateId and version.
int actingBlockLength = MessageHeader.BlockLength;
int actingVersion = MessageHeader.Version;

bufferOffset += MessageHeader.Size;
// now we decode the message
CarExample.Decode(Car, directBuffer, bufferOffset, actingBlockLength, actingVersion);
```
You have specified a schema version for the MessageHeader using `Car.SchemaVersion` before you know that it is a `Car` message contained by examining the MessageHeader's `TemplateId`
SchemaVersion has a common value for each compiled class, and so should be a value in the generated namespace rather than specific to each class.

As an alternative the SchemaVersion static field could be removed from all the generated classes for the individual message types and moved to the MessageHeader class.

mjpt777:
All the flyweights now have a constant members for schema id and version so you can now do:

```
    MessageHeader.Wrap(directBuffer, bufferOffset, MessageHeader.SbeSchemaVersion);
```

PR #779, opened 2020-05-05
[C#] Add SchemaVersion to MessageHeader

szehetner:
Adresses #728 by adding the SchemaVersion to the MessageHeader and all composite types (in line with the generated Java code).


mjpt777:
I took a different approach with this commit.  https://github.com/real-logic/simple-binary-encoding/commit/47226c4f668d83c163f45fa1821224e1ed250647

PR #780, opened 2020-05-13
Indexed fields as spans

gatesy:
This PR exposes an array field as a `ReadOnlySpan`.

This was motivated by the need to store a GUID in a field. The natural way to do this is using a field of 16 bytes, which means the way to extract and encode it requires copying to an intermediate array. The C# GUID struct provides a `ReadOnlySpan` constructor, allowing us to create a GUID without copying to an intermediate byte buffer, e.g.:
```
var guid = new Guid(message.Guid);
```
and for encoding store:
```
var guidBytes = Guid.NewGuid().ToByteArray();
```
then set it using:
```
message.Guid = new ReadOnlySpan<byte>(guidBytes);
```
I've also added an `AsSpan()` method, to support writing directly from a Guid object to the underlying array. This allows you to write a Guid like this:
```
guid.TryWriteBytes(message.GuidAsSpan());
```


mjpt777:
@billsegall Do you have any views on this?

billsegall:
I think it's a useful helper function. I see no compatibility issues.

gatesy:
Could we get this change merged up and a new version of the sbe-tool NuGet package published? Is there anything I can help with to get this done?

billsegall:
A new nuget release is probably due. I'll do that post merge.

Bill.

billsegall:
I've done a nuget release as 1.17.1. Consistent versioning across all backend languages is a little tricky.

PR #784, opened 2020-05-24
Fixes a cpp warning.

lygstate:
The uint64_t of messageHeader::encodedLength() large than (%1$s), so we need cast to it


mjpt777:
I think this approach is incorrect. The original signature should have returned a `uint64_t` given it used `encodedLength()`. A more correct approach would be to return `uint64_t`

lygstate:
%1$s -> uint64_t ? also acceptable.

tmontgomery:
I agree with @mjpt777 .

lygstate:
Update with uint64_t

PR #785, opened 2020-05-24
I suggest do not alter the case in C generator, maintain consistence with C++/Java/C#/XML

lygstate:
If we change the case.

Support we have something like UUID
then it turn out to be uUID(That's weird.)


mjpt777:
UUID is a bad name. All caps abbreviations are difficult to tell apart.

PR #787, opened 2020-05-24
Fixes compiling warning on MSVC by cast the type properly

lygstate:


tmontgomery:
I think these should probably use static_cast or reinterpret_cast as appropriate.

PR #792, opened 2020-05-26
Windows CI fixes.

lygstate:


tmontgomery:
We do not want to remove the external project add and do the extract method in this. And do not add to our CI. @vyazelenko will be looking into this later.

PR #793, opened 2020-05-28
constexpr -> SBE_CONSTEXPR

lygstate:


mjpt777:
This introduces more warnings to the build rather than reducing.

lygstate:
What warning it cause? Old compiler doesn't have constexpr

mjpt777:
> What warning it cause?
https://github.com/real-logic/simple-binary-encoding/pull/793/checks?check_run_id=715960514

lygstate:
There is another fixes about the warning
The warning was hidden by constexpr

PR #802, opened 2020-07-09
Upgrade csharp benchmark, samples and tests console app projects to u…

alexisatkinson:
…se dotnet core 3.1 (LTS) as opposed to the end of life dotnet core 2.2 (which less people will have installed)


alexisatkinson:
Closing due to CI failure.  Will replace .net core 2.2 dependency with .net core 3.1 in github actions.

PR #814, opened 2020-08-24
C# auto-generated header for generated c# codec files

mhoogveld:
Adding the <auto-generated> header to generated .cs files will cause them to be ignored by Visual Studio's code-inspection, which will speed up the IDE.
Also changed to file-extension from *.cs to *.g.cs, which has exactly the same result in Visual Studio, but can also be used by other tools (e.g. Git's .gitignore or Visual Studio's .editorconfig)

The *.g.cs extension is mentioned in various places and is used since VS 2008 or earlier, but there does not seem to be any official documentation.
The best reference I found regarding the extension is https://developers.google.com/protocol-buffers/docs/reference/csharp-generated


mjpt777:
@billsegall Do you have a view on this?

billsegall:
I can see it makes sense but I have no experience on the usage of .g.cs.

This pull request seems not to change .gitignore files ot VS projects. Is
it complete?

I won't get any time to look further until next week,

Bill.

mjpt777:
Please ensure the build passes all tests before submitting.

mhoogveld:
@billsegall I'm not sure what .gitignore file(s) you refer to.
The .gitignore in the root directory already includes all generated c# code in /csharp/sbe-generated by ignoring directories themselves or *.cs files in them (which also include the *.g.cs files of course)

@mjpt777 If I'm correct, the failing test already fails in the commit in the master on which this branch is based.
Would you like me to to rebase this branch onto another existing commit?

PR #692, opened 2019-06-25
Improved internal offset/pointer usage in generated C# code

mhoogveld:
* Improved internal pointer usage in generated C# code regarding Groups by making it possible to call WrapForEncode and WrapForDecode twice. This makes it possible to read a message twice which is especially useful for logging the entire message after errors or when unexpected content is found.
* Enabled CSharp code generator in the sbe-tool.

Issue #693, opened 2019-07-01
[C#] Property accesses for groups have side effects


JulianBirch:
The generated C# for groups updates the parent message limit. This, sadly, is more than a conceptual complaint. It has the following effects:

- Anyone putting the properties in their debugger will break their code in a hard to find way.
- Anyone with "Implicit Evaluation" switched on in visual studio can break their running code just by mouse-overing the expression.

Obviously, fixing this properly would involve removing these properties and replacing with functions, which would break existing code. Adding the `[DebuggerHidden]` attribute to the generated code might ameliorate the worst side-effects, though.

PR #815, opened 2020-08-24
Improved internal offset/pointer usage in generated C# code (retry)

mhoogveld:
Retry of pull-request regarding issue #692.

* Improved internal pointer usage in generated C# code regarding Groups by making it possible to call WrapForEncode and WrapForDecode more than once. This makes it possible to read a message twice which is especially useful for logging the entire message after errors or when unexpected content is found.
* Enabled CSharp code generator in the sbe-tool.


mjpt777:
Please ensure the build passes all tests before submitting.

mhoogveld:
I noticed the failing checks. I'll fix them and update the branch

mhoogveld:
Although this solves issue #693, it breaks reading groups with variable length data in them.
I cannot see an easy solution which solves both. 
I'm closing this request.

PR #820, opened 2020-10-26
[Java] Generate appendTo which skips decoder allocation

ariksher-nex:
using `decoderForDisplay` as private final member and reusing it , in order to avoid memory allocation (of the decoder)

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/java/JavaGenerator.java
@@ -3234,6 +3234,8 @@ public class JavaGenerator implements CodeGenerator
 
         appendToString(sb, INDENT);
         sb.append('\n');
+        append(sb, INDENT, "private final " + decoderName + " decoderForDisplay = new " + decoderName + "();");
+        sb.append('\n');
         append(sb, INDENT, "public StringBuilder appendTo(final StringBuilder builder)");
         append(sb, INDENT, "{");
         append(sb, INDENT, "    if (null == buffer)");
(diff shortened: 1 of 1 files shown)


mjpt777:
This would increase the allocation footprint for those who do not call toString so I'm not inclined to merge it.

vyazelenko:
How about having a lazy-initialized field for the decoder. This way the overhead should be max 4 (8) bytes for the reference for those not using the `toString`.

mjpt777:
Or someone can have their own copy of the decoder and use it for that purpose. The toString is useful for debugging and not for performance.

vyazelenko:
Agree.

ariksher-nex:
> 
> 
> Or someone can have their own copy of the decoder and use it for that purpose. The toString is useful for debugging and not for performance.

if want to maintain the decoder from outside so still need to generate a toString method which would accept it as parameter, should i propose such?

ariksher-nex:
> 
> 
> How about having a lazy-initialized field for the decoder. This way the overhead should be max 4 (8) bytes for the reference for those not using the `toString`.

can implement such as well of course, please tell me what you prefer

mjpt777:
There is already a [toString(StringBuilder](https://github.com/real-logic/agrona/blob/master/agrona/src/main/java/org/agrona/sbe/MessageDecoderFlyweight.java#L42) on the decoder which can be used. If you care about allocation then you should be doing that. Another option is to write your own OTF decoder similar to the [JsonPrinter](https://github.com/real-logic/simple-binary-encoding/blob/master/sbe-tool/src/main/java/uk/co/real_logic/sbe/json/JsonPrinter.java).

ariksher-nex:
> 
> 
> There is already a [toString(StringBuilder)](https://github.com/real-logic/agrona/blob/master/agrona/src/main/java/org/agrona/sbe/MessageDecoderFlyweight.java#L42) on the decoder which can be used. If you care about allocation then you should be doing that. Another option is to write your own OTF decoder similar to the [JsonPrinter](https://github.com/real-logic/simple-binary-encoding/blob/master/sbe-tool/src/main/java/uk/co/real_logic/sbe/json/JsonPrinter.java).

@mjpt777 yeah, i want to avoid allocation during runtime.
I'm not familiar with OTF feature/code/scope, and I don't have the time to learn about it.

Is it acceptable to go with the approach of lazy initialization of `decoderForDisplay` as @vyazelenko suggested?

mjpt777:
This approach has a minimal impact on allocation as the majority is in the StringBuilder and the String which results. You need to keep another instance of a decoder and use that with an instance of a StringBuilder you can reuse. If you are unclear with how to do that or use the OTF then we can offer consulting.

ariksher-nex:
I appreciate your will to consult about OTF, not these days, thanks :-)
I will implement your idea about maintaining my own decoder. 
Thanks.

PR #825, opened 2020-12-13
[Go] Fix for decoding into a struct with pre-allocated slices.

ethanf:
This change allows you to call `Decode()` on an sbe struct where the struct has pre-allocated slices. Before, the code would allocate a new slice if the capacity on the existing slice was less than the number of groups in the message. If the old slice had a greater capacity, it wouldn't truncate the number of elements to the new length. The same issue with vardata fields.


ethanf:
@billsegall can you have a look?

billsegall:
I like it. It's a neat fix.

PR #841, opened 2021-04-01
[C#] Convert benchmarks to BenchmarkDotNet

szehetner:
As mentioned in https://github.com/real-logic/simple-binary-encoding/pull/840, this converts the C# benchmark project to use BenchmarkDotNet instead of the handcrafted benchmark logic.

Notable changes:
* The target framework (for the benchmark project only) was changed from .NET Framework 4.5 to 4.6.1 as this is the minimum required by BenchmarkDotNet (needs to be netstandard 2.0 compatible).
* The output assembly was renamed to sbe-benchmarks (BenchmarkDotNet requires the assembly to have the same name as the .csproj file)

Benchmarks:
* MarketDataBenchmark is basically the same as the existing one, only the BenchmarkState class was merged into it.
* CarBenchmark is based on the car example, covering more features.


mjpt777:
Note there are a lot of warnings after merging this PR in the benchmarks. The assigned values probably need to be consume to make the benchmark valid.

Issue #844, opened 2021-05-10
[Go] SbeTool generator FormatFlagsConversionMismatchException


neomantra:
I found a fragility in the SbeTool generator for Golang.   I got it to a simple reproducible case.

 Process the following schema with `java -Dsbe.target.language=Golang -jar sbe-all-1.22.0.jar golang_bug.xml` and it will succeed... it will also succeed on all the other generators:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<messageSchema package="gobug" id="1" semanticVersion="1.0" description="Go Bug" byteOrder="littleEndian">
    <types>
        <composite name="Comp1" description="A composite">
            <type name="lmn"  primitiveType="uint16"/>
            <type name="wxyz" primitiveType="uint16"/>
        </composite>
        <composite name="messageHeader" description="Message identifiers and length of message root">z
            <type name="blockLength" primitiveType="uint16"/>
            <type name="templateId"  primitiveType="uint16"/>
            <type name="schemaId"    primitiveType="uint16"/>
            <type name="version"     primitiveType="uint16"/>
            <ref  name="o"           type="Comp1"/>
            <type name="abcd"       primitiveType="uint16"/>
            <ref  name="p"           type="Comp1"/>
        </composite>
    </types>
    <message name="barmsg" id="4">
        <field name="header"   id="1" type="messageHeader"/>
    </message>
</messageSchema>
```

Change `<type name="abcd"` to `<type name="abcde"` and it will fail with this exception:

```
Exception in thread "main" java.util.FormatFlagsConversionMismatchException: Conversion = s, Flags = 0
        at java.base/java.util.Formatter$FormatSpecifier.failMismatch(Formatter.java:4422)
        at java.base/java.util.Formatter$FormatSpecifier.checkBadFlags(Formatter.java:3151)
        at java.base/java.util.Formatter$FormatSpecifier.checkGeneral(Formatter.java:3109)
        at java.base/java.util.Formatter$FormatSpecifier.<init>(Formatter.java:2870)
        at java.base/java.util.Formatter.parse(Formatter.java:2713)
        at java.base/java.util.Formatter.format(Formatter.java:2655)
        at java.base/java.util.Formatter.format(Formatter.java:2609)
        at java.base/java.lang.String.format(String.java:3292)
        at uk.co.real_logic.sbe.generation.golang.GolangGenerator.generateTypeBodyComposite(GolangGenerator.java:2019)
[...]

PR #846, opened 2021-05-11
[Go] GolangGenerator improvements

neomantra:
Adds function `generateWhitespace` which uses the same String.format
technique, but validates the input.  It also always generates at least
one space.

This change should only affect the whitespace of the output.  It does
not attempt to change the logic that might be the underlying cause of #844.


billsegall:
This seems sensible.

Bill

Issue #849, opened 2021-05-13
[Go] GolangGenerator nested composite types are malformed


neomantra:
Given the following schema:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<sbe:messageSchema xmlns:sbe="http://fixprotocol.io/2016/sbe"
    package="issue849"
    id="1"
    version="0"
    semanticVersion="1.0"
    description="test case 849 20210512"
    byteOrder="littleEndian">
    <types>
        <composite name="Comp1" description="Comp1">
            <type name="abc"  primitiveType="uint32"/>
            <type name="wxyz" primitiveType="uint32"/>
        </composite>
        <composite name="messageHeader" description="MH">
            <type name="blockLength" primitiveType="uint16"/>
            <type name="templateId"  primitiveType="uint16"/>
            <type name="schemaId"    primitiveType="uint16"/>
            <type name="version"     primitiveType="uint16"/>
            <ref  name="c1"          type="Comp1"/>
            <type name="lmn"         primitiveType="uint32"/>
            <type name="opq"         primitiveType="uint32"/>
        </composite>
    </types>
    <sbe:message name="barmsg" id="4">
        <field name="header" id="1" type="messageHeader"/>
    </sbe:message>
</sbe:messageSchema>
```

The generated nested types have too many fields (`Lmn`):

```
type MessageHeaderComp1 struct {
    Abc  uint32
    Wxyz uint32
    Lmn  uint32
}
```

And compilation brings up these errors:
```
./MessageHeaderComp1.go:53:7: m.LmnInActingVersion undefined (type *MessageHeaderComp1 has no field or method LmnInActingVersion)
./MessageHeaderComp1.go:54:12: m.LmnNullValue undefined (type *MessageHeaderComp1 has no field or method LmnNullValue)
./MessageHeaderComp1.go:74:6: m.LmnInActingVersion undefined (type *MessageHeaderComp1 has no field or method LmnInActingVersion)
./MessageHeaderComp1.go:75:15: m.LmnMinValue undefined (type *MessageHeaderComp1 has no field or method LmnMinValue)
./MessageHeaderComp1.go:75:42: m.LmnMaxValue undefined (type *MessageHeaderComp1 has no field or method LmnMaxValue)
./MessageHeaderComp1.go:76:69: m.LmnMinValue undefined (type *MessageHeaderComp1 has no field or method LmnMinValue)
[...]

PR #850, opened 2021-05-13
[Go] Fix tokens passed to generateComposite (#849)

neomantra:
Too many tokens were passed to `generateComposite` causing weird
errors like(#844 and #849).

Also adds a `gofmt` fix.


billsegall:
That check failure looks transient?

This looks like a good fix to a clear bug in <ref> handling.

And it's always great to get a regression test case with the PR.

neomantra:
Yes, it looks like that one CI job failed with “502 bad gateway”, some network hiccup during setup.

PR #870, opened 2021-09-28
Add sbeLimit methods.

mikeb01:
To work with the updated flyweight methods in Agrona.


mjpt777:
This PR is no longer necessary.

PR #872, opened 2021-11-03
Generate stubs for non used composites

xdmiodz:
Currently the `sbe-tool` generates stubs only for composites which are being used in a `sbe:message`. For my purposes I need generate stubs for all the composites, not only referenced in the messages. This patch allows to do this.


mikeb01:
Looks like some build errors.  You will need to regenerate the IR codecs.  Instructions are in the build failure.

mjpt777:
Closing due to no progress.

Issue #875, opened 2021-11-10
Deprecated messages should result in their en/decoders marked with @Deprecated


ratcashdev:
given a definition like
```xml
<sbe:message name="MyDeprecatedMessage" id="123"
    description="old message that was deprecated in version 45" deprecated="45">
```
its respective java encoder and decoder should be marked with the `@Deprecated` annotation so that users get compile time warnings when still using old messages.

Happy to submit a PR, if this issue and objectives are accepted.

mjpt777:
I'd be happy to review an PR on this.

PR #876, opened 2021-11-10
Mark classes of deprecated messages with @Deprecated.

ratcashdev:
Implementation of #875

Wondering is something similar would make sense for fields?

### DIFF
--- a/sbe-tool/src/test/resources/since-deprecated-test-schema.xml
+++ b/sbe-tool/src/test/resources/since-deprecated-test-schema.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <sbe:messageSchema xmlns:sbe="http://fixprotocol.io/2016/sbe"
-                   package="since-deprecated"
+                   package="since.deprecated"
                    id="108"
                    version="4"
                    semanticVersion="5.2"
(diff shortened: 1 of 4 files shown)


sbe-tool/src/test/resources/since-deprecated-test-schema.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <sbe:messageSchema xmlns:sbe="http://fixprotocol.io/2016/sbe"
-                   package="since-deprecated"
+                   package="since.deprecated"
```
vyazelenko, line 3: This change breaks the C# build (see https://github.com/real-logic/simple-binary-encoding/runs/4171818266?check_suite_focus=true):
```
Error: SinceDeprecatedTests.cs(72,49): error CS0103: The name 'SinceDeprecated' does not exist in the current context [/home/runner/work/simple-binary-encoding/simple-binary-encoding/csharp/sbe-tests/sbe-tests.csproj]
Error: SinceDeprecatedTests.cs(73,49): error CS0103: The name 'SinceDeprecated' does not exist in the current context [/home/runner/work/simple-binary-encoding/simple-binary-encoding/csharp/sbe-tests/sbe-tests.csproj]
    0 Warning(s)
    42 Error(s)
```

ratcashdev: I see, sorry. Then I create a separate XML for this.

PR #860, opened 2021-07-21
New Rust code generator

mward:
This PR provides full SBE code generation support for the Rust programming language.

- generates 100% safe rust code
- the rust code generated has zero dependencies on any external rust libs
- both `lib.rs` and `Cargo.toml` files are generated so projects can easily add codec as a dependency
- Fixes #813 

#### Testing
```
# following will write generated code to "./generated/rust"
./gradlew generateRustCodec

# assuming you have "rust" and "cargo" already installed
cd ./rust
cargo test
```

PR #877, opened 2021-11-12
Generating Rust code should support multiple schema files with same package name

mward:
The generated `lib.rs` file now only contains common sbe code, all generated enums, bitSets, composites, etc. will be written to their own file.


mward:
@joakimthun - this PR should fix the issue you discovered.

https://github.com/real-logic/simple-binary-encoding/pull/860#issuecomment-951997233

joakimthun:
LGTM - tested the changes locally and I know see all enums getting generated and all types exported.

joakimthun (approved):

PR #888, opened 2022-01-31
C++ code generation: check composite and bitset version

greenstein:
Previously, the C++ generated code did not check the version before writing composite and bitset fields to the output stream. This leads to undefined behavior. I've updated the generated code to first check if the field is in the acting version before writing it to the stream.


tmontgomery:
I'm good with this assuming it passes the build.

PR #921, opened 2022-12-18
Fixes for Rust Composites

adam-talos:
Fixed several issues:
1: LibRsDef explicitly use crate:: to disambiguate from built in mods like bool
2: Added acting_version field to Composite structs to fix compilation errors when using Composite structs.  This is an incomplete implementation because the parent doesn't pass the acting_version to the composite because you need to change the signature of wrap(parent, offset) to include the acting_version, so this version just ensures that if the acting_version isn't set on the composite, it disregards the version check.
3: fixed primitiveArrayDecoder to return an empty array of the right size if less than version required.
4: If a name of an element clashes with a rust type, prefix it with r#, which is the preferred way of dealing with clashes with keywords

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java
@@ -542,7 +542,7 @@ public class RustGenerator implements CodeGenerator
                 decoderName,
                 decoderTypeName);
 
-            indent(sb, level + 1, "if self.acting_version < %d {\n", fieldToken.version());
+            indent(sb, level + 1, "if self.acting_version > 0 && self.acting_version < %d {\n", fieldToken.version());
             indent(sb, level + 2, "return Either::Left(self);\n");
             indent(sb, level + 1, "}\n\n");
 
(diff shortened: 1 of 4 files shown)


mikeb01:
This is failing the CI tests, please could you look into the failures and make the necessary corrections.

vyazelenko:
@adam-talos The build fails with the CheckStyle errors:
```
$ ./gradlew
[ant:checkstyle] [ERROR] simple-binary-encoding/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java:942: Line is longer than 120 characters (found 128). [LineLength]
[ant:checkstyle] [ERROR] simple-binary-encoding/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java:957: Line is longer than 120 characters (found 128). [LineLength]
[ant:checkstyle] [ERROR] simple-binary-encoding/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java:976: Line is longer than 120 characters (found 124). [LineLength]
```

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java
```diff
@@ -790,7 +790,7 @@ private static void generatePrimitiveRequiredDecoder(
 
         if (fieldToken.version() > 0)
         {
-            indent(sb, level + 1, "if self.acting_version < %d {\n", fieldToken.version());
+            indent(sb, level + 1, "if self.acting_version > 0 && self.acting_version < %d {\n", fieldToken.version());
```
SeanKim, line 793: I wonder why `self.acting_version > 0` condition is added.
Is `self.acting_version < %d` is not enough for version checking?

SeanKim: I figured out why `self.acting_version > 0` is required.
acting_version is set to 0 when there is no acting_version requirement.

vyazelenko:
@adam-talos This PR has merge conflicts and failing tests. Are you going to fix it or shall we close it?

PR #939, opened 2023-05-10
CppGenerator: operator<< take messages as const-ref

SBAM:
Dear all,

Following commit 1973c86f84b190d5b2507786e6445970c390ea62, using generated operator<< would require a mutable ref message.
This is unnecessary and can be problematic.
For example when using wrappers such as fmt::streamed ( https://fmt.dev/latest/api.html#std-ostream-support ) that enforce const-correctness.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -2564,7 +2564,7 @@ public class CppGenerator implements CodeGenerator
         new Formatter(sb).format("\n" +
             "template<typename CharT, typename Traits>\n" +
             "friend std::basic_ostream<CharT, Traits> & operator << (\n" +
-            "    std::basic_ostream<CharT, Traits> &builder, %1$s &_writer)\n" +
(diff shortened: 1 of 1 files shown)


SBAM:
Hello @ksergey 

Apologies if I misunderstood your comment but method **generateGroupDisplay** was not modified. Dumping a Group should still require a mutable ref, should it not ?

SBAM:
Dumping directly a Group indeed mutates it since 1973c86f84b190d5b2507786e6445970c390ea62. I assume that people who did that copied Group first.

However, dumping a regular message copies initial data and passes to cascading groups dumpers that copy.

```cpp
template<typename CharT, typename Traits>
friend std::basic_ostream<CharT, Traits> & operator << (
    std::basic_ostream<CharT, Traits> &builder, const MDIncrementalRefreshBook46 &_writer)
{
    MDIncrementalRefreshBook46 writer(
        _writer.m_buffer,
        _writer.m_offset,
        _writer.m_bufferLength,
        _writer.m_actingBlockLength,
        _writer.m_actingVersion);

    builder << '{';
```
Are you suggesting to revert 1973c86f84b190d5b2507786e6445970c390ea62 completely ?

SBAM:
I still can't quite grasp how adding that const will make current implementation work less well.
Does it generate issues that are not already there ?

Issue #765, opened 2020-03-19
[Go] no zero-copy?


asido:
I generated Go code and see there is no `Wrap()` method. Instead, there is `Encode(io.Reader)`, which fetches and assigns all values to public fields in the model. Is there no zero-copy alternative in Go, that would allow me to read the data in-place?

neomantra:
This is correct.  The current Go implementation takes raw SBE message and creates a fully-formed Golang struct from it and thus every field is marshaled.

The reasoning behind this is described here:
https://github.com/real-logic/simple-binary-encoding/wiki/Golang-User-Guide#golang-idioms

This  design prioritizes easy-to-use and hard-to-misuse over performance.    It should be possible to make a zero-copy / flyweight generator if someone were up to the task.

matthewdowney:
@ethanf curious if your impl ended up looking promising?

ethanf:
> @ethanf curious if your impl ended up looking promising?

I've updated https://github.com/ethanf/simple-binary-encoding/tree/go-flyweight with the latest code. It's code complete and mostly works, but there's still work to do updating the tests unfortunately.

ethanf:
https://github.com/ethanf/simple-binary-encoding/tree/go-flyweight has passing tests now, does anyone have a preference for how to switch between the flyweight and struct versions of the generated code? Is a system property that switches between the generators acceptable?

neomantra:
Thanks for your contribution!   I only started looking at it and will take it for a spin in a few weeks.

I do think that using system properties like that is consistent with other SBE / Aeron tool options.

PR #951, opened 2023-08-08
[Go] SBE implementation in Go using flyweights

ethanf:
See #765. To generate flyweights instead of structs, set the following flag:

```
sbe.go.generate.generate.flyweights=true
```

Includes OTF and JsonPrinter implementations for parity with other implementations.


vyazelenko:
@ethanf There are merge conflicts which prevent this PR from being merged.

PR #960, opened 2023-11-02
[C++] Generate DTOs for non-perf-sensitive usecases.

ZachBray:


ZachBray:
Combining with C# changes, as both rely on PBT infrastructure.

PR #963, opened 2023-11-27
[Java] Support camelCase enum type identifiers for value refs

pliard:
The type identifier was previously used as-is (e.g. `carFilterType`) in
the generated code for uses of `valueRef` while generated Java enums
always start with a capital letter (`CarFilterType`).

See the added test in `ValueRefWithLowerCaseEnum.java`.


mjpt777:
Please provide more detail as this would be a breaking change for existing code.

pliard:
> Please provide more detail as this would be a breaking change for existing code.

Thanks for the quick reply. Please take another look. I've added a test to show the issue.

PR #965, opened 2023-12-07
feat: updated gradle to 8.5 for jdk21 support

klittlepage:
* updated gradle to 8.5
* explicitly declared org.junit.platform:junit-platform-launcher dependency for gradle (implicit deps deprecated in 8.5, error in 9.X)
* handled missing directory for sbe-all (projects without an existing directory deprecated in 8.5, error in 9.X)


damondd:
why not merged yet after 6 months ???

vyazelenko:
@damondd We use the same version of Gradle across Aeron, SBE and Agrona. We cannot upgrade Gradle without breaking some plugins whose newer version require JDK 11+. Therefore until we move past JDK 8 we won't be upgrading Gradle.

That being said, we are testing with JDK 17, 21 and 23-ea (Agrona only). This requires specifying `BUILD_JAVA_VERSION` and `BUILD_JAVA_HOME`, e.g.:
```bash
$ BUILD_JAVA_VERSION=21 BUILD_JAVA_HOME=/usr/lib/jvm/zulu-21 ./gradlew
```

Issue #984, opened 2024-04-05
Rust: invalid codegen for groups with fields of type with 'sinceVersion'


mward:
This is similar to #972 but specific to simple `type` with sinceVersion > 0

```xml
<sbe:messageSchema
    xmlns:sbe="http://fixprotocol.io/2016/sbe"
    package="example"
    id="123"
    version="3"
    semanticVersion="5.0.0"
    byteOrder="littleEndian">
  <types>
    <type name="String4" length="4" primitiveType="char" semanticType="String"/>
    <type name="String5" length="5" primitiveType="char" sinceVersion="2" semanticType="String"/>
    <type name="String6" length="6" primitiveType="char" sinceVersion="3" semanticType="String"/>
    <composite name="groupSize" description="Repeating group dimensions" semanticType="NumInGroup">
      <type name="blockLength" primitiveType="uint16"/>
      <type name="numInGroup" primitiveType="uint8"/>
    </composite>
    <composite name="messageHeader">
      <type name="blockLength" primitiveType="uint16"/>
      <type name="templateId" primitiveType="uint16" description="Template ID used to encode the message."/>
      <type name="schemaId" primitiveType="uint16" description="ID of the system publishing the message."/>
      <type name="version" primitiveType="uint16" description="Schema version."/>
    </composite>
  </types>
  <sbe:message name="SimpleMessage" id="1">
    <field name="id" id="2" type="uint16" semanticType="Int"/>
    <group name="MyGroup" id="3" blockLength="15" dimensionType="groupSize">
      <field name="f1" id="4" type="String4" offset="0" semanticType="String"/>
      <field name="f2" id="5" type="String5" offset="4" semanticType="String"/>
      <field name="f3" id="6" type="String6" offset="9" semanticType="String"/>
    </group>
  </sbe:message>
</sbe:messageSchema>
```

rust compiler complains with: 
```
error[E0609]: no field `acting_version` on type `&simple_message_codec::decoder::MyGroupDecoder<P>`
   --> src/simple_message_codec.rs:382:21
    |
382 |             if self.acting_version > 0 && self.acting_version < 2 {
    |                     ^^^^^^^^^^^^^^ unknown field
    |
    = note: available fields are: `parent`, `block_length`, `count`, `index`, `offset`
```
I have a fix that will resolve this, will submit pull request

mward:
Submitted pull request: #985

vyazelenko:
PR merged.

PR #985, opened 2024-04-05
[Rust] Updated code generator to resolve Issue #984

mward:


mward:
looks like the GCC 12 compiler failed to install on Ubuntu 22.04 for some reason (github container issue?).  I had similar issue on my fork with GCC 11 and Ubuntu 22.04.  On my fork I was able to re-run that step and it passed, but I don't have access to try that here

vyazelenko:
@mward I've seen issues with GCC 12 on Aeron repo as well.

Issue #988, opened 2024-04-10
[Java] Precedence checks are false negative when decoding a older version message


zyulyaev:
Decoders generated with `sbe.generate.precedence.checks=true` throw `IllegalStateException` when message from a previous version, that is not mentioned in the message schema, is decoded in the correct order.

For example the following schema:
```
     <sbe:message id="1000" name="BugShowcase">
        <field id="1" name="field" type="int64"/>
        <group id="2" name="group" sinceVersion="3">
        </group>
        <data id="3" name="data" type="varStringEncoding"/>
    </sbe:message>
```

results in the follow state machine:
```mermaid
graph TD
    NOT_WRAPPED-->|"wrap(version=0)"|V0_BLOCK
    V0_BLOCK -->|"field(?)"|V0_BLOCK
    V3_BLOCK-->|"field(?)"|V3_BLOCK
    V0_BLOCK-->|"dataLength()"|V0_BLOCK
    V3_GROUP_1_BLOCK-->|"dataLength()"|V3_GROUP_1_BLOCK
    V3_GROUP_DONE-->|"dataLength()"|V3_GROUP_DONE
    V0_BLOCK-->|"data(?)"|V0_DATA_DONE
    V3_GROUP_1_BLOCK-->|"data(?)"|V3_DATA_DONE
    V3_GROUP_DONE-->|"data(?)"|V3_DATA_DONE
    NOT_WRAPPED-->|"wrap(version=3)"|V3_BLOCK
    V3_BLOCK-->|"groupCount(0)"|V3_GROUP_DONE
    V3_BLOCK-->|"groupCount(>0)"|V3_GROUP_N
    V3_GROUP_N_BLOCK-->|"group.next()\n  where count - newIndex > 1"|V3_GROUP_N_BLOCK
    V3_GROUP_N-->|"group.next()\n  where count - newIndex > 1"|V3_GROUP_N_BLOCK
    V3_GROUP_N_BLOCK-->|"group.next()\n  where count - newIndex == 1"|V3_GROUP_1_BLOCK
    V3_GROUP_N-->|"group.next()\n  where count - newIndex == 1"|V3_GROUP_1_BLOCK
    V3_GROUP_N_BLOCK-->|"group.resetCountToIndex()"|V3_GROUP_DONE
    V3_GROUP_1_BLOCK-->|"group.resetCountToIndex()"|V3_GROUP_DONE
[...]

PR #989, opened 2024-04-11
[Java, C#, C++] Fix precedence checks to account for non-contiguous versions.

ZachBray:
Relates to issue #988.

An SBE message might not change in every version of a schema. For example, another message might change.

Previously, the field precedence checking model would expect an exact match for `actingVersion` in `wrap` to enter one of its initial states. However, it is only aware of the versions in which a message schema has changed. Therefore, it was possible for `actingVersion` not to match any of these versions, in which case the initial state that represented a codec wrapped around the latest known version of the format was picked.

Now, we do _not_ expect an exact match in `wrap`. Instead, we select the "best" match. For example, if a codec was changed in v1 and v3, and the acting version is v2, we will decode using v1. However, if the acting version is v4, we would decode as v3.

### DIFF
--- a/csharp/sbe-tests/FieldAccessOrderCheckTests.cs
+++ b/csharp/sbe-tests/FieldAccessOrderCheckTests.cs
@@ -3159,6 +3159,25 @@ namespace Org.SbeTool.Sbe.Tests
             var exception = Assert.ThrowsException<InvalidOperationException>(encoder.CheckEncodingIsComplete);
             StringAssert.Contains(exception.Message, $"Not fully encoded, current state: {expectedState}");
         }
+        
+        [TestMethod]
+        public void AllowsSkippingFutureGroupWhenDecodingFromVersionWithNoChangesInMessage()
+        {
+            var encoder = new AddGroupBeforeVarDataV0()
+                .WrapForEncodeAndApplyHeader(_buffer, Offset, _messageHeader);
+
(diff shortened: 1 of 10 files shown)


vyazelenko:
@ZachBray Looks but the C# tests fail.

vyazelenko (approved):

vyazelenko:
@ZachBray 
```C#
Error Message:
   Test method Org.SbeTool.Sbe.Tests.FieldAccessOrderCheckTests.AllowsOldDecoderToSkipAddedPrimitiveFieldBeforeGroup threw exception: 
Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException: Method Debug.Fail failed with '
', and was translated to Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException to avoid terminating the process hosting the test.
  Stack Trace:
      at Org.SbeTool.Sbe.Tests.FieldAccessOrderCheckTests.ModifyHeaderToLookLikeVersion1() in /home/runner/work/simple-binary-encoding/simple-binary-encoding/csharp/sbe-tests/FieldAccessOrderCheckTests.cs:line 3197
   at Org.SbeTool.Sbe.Tests.FieldAccessOrderCheckTests.AllowsOldDecoderToSkipAddedPrimitiveFieldBeforeG
```

PR #991, opened 2024-04-24
Help decoding multiple messages using rust by resolving #987

### LINKED ISSUE #987: [Rust] cumulate offset of header and message_header_codec::ENCODED_LENGTH in generated Decoder::header method
Code generator for rust would generate a `header` method for Decoder (see following)
```rust
impl<'a> DonotUsePlaceHolderDecoder<'a> {
        pub fn wrap(
            mut self,
            buf: ReadBuf<'a>,
            offset: usize,
            acting_block_length: u16,
            acting_version: u16,
        ) -> Self {
            let limit = offset + acting_block_length as usize;
            self.buf = buf;
            self.initial_offset = offset;
            self.offset = offset;
            self.limit = limit;
            self.acting_block_length = acting_block_length;
            self.acting_version = acting_version;
            self
        }

        #[inline]
        pub fn encoded_length(&self) -> usize {
            self.limit - self.offset
        }

        pub fn header(self, mut header: MessageHeaderDecoder<ReadBuf<'a>>) -> Self {
            debug_assert_eq!(SBE_TEMPLATE_ID, header.template_id());
            let acting_block_length = header.block_length();
            let acting_version = header.version();

            self.wrap(
                header.parent().unwrap(),
                message_header_codec::ENCODED_LENGTH,
                acting_block_length,
                acting_version,
            )
        }
}
```

The `header` method take a header decoder and call `wrap` method, passing `message_header_codec::ENCODED_LENGTH` as offset.

Considering a header decoder may be wraped with a none-zero offset `header_off`, `header` method should wrap with `header_off + message_header_codec::ENCODED_LENGTH`.  To support this functionality, a MessageHeaderDecoder should expose its offset via, say `get_limit()` or something else.

wbprime:
Closes #987 

Say multiple messages were encoded into a continuous byte buffer.  When trying to decode the first message, we might first use a `MessageHeaderDecoder` to get its templateId at offset=0, then decode its body using correspoded message decoder at offset+=headerLength.  This is ok for current message decoder's `header` method with an implicit header offset=0.  But when trying to decoder the second and following messages, the `MessageHeaderDecoder` step is ok however the body decoding step using `header` method would fetch bytes always from the first message body, thus is not expected.

This unexpected result occurs only when using message decoder's `header` method to gain ownership.  Use message decoder's `wrap` method behaves as expected.

This pr is just a simple workaround, simple but works (see tests/issue_987_test.rs).


mward:
Thanks @wbprime, I think this looks good.  Can someone apply this PR?

Issue #992, opened 2024-04-25
[C#] ToString() affects the message state


amibar:
The generated ToString() method of a message affects the state of the message. 
Since the debugger calls to ToString() to show its data in the watches, it messes up with the state of the program.

Moreover, quoting from the `ToString()` documents "Notes to Inheritors": https://learn.microsoft.com/en-us/dotnet/api/system.object.tostring
* Your `ToString()` override should have no observable side effects to avoid complications in debugging. For example, a call to the `ToString()` method should not change the value of instance fields.

PR #993, opened 2024-04-25
[C#] Avoid ToString() affect the message internal state

amibar:
Hi,

This PR intends to fix [Issue 992](https://github.com/real-logic/simple-binary-encoding/issues/992).

The current generate ToString() implementation calls to `this.BuildString()` which change `this` state.

```
public override string ToString()
{
    var sb = new StringBuilder(100);
    this.BuildString(sb);
    return sb.ToString();
}
```

To avoid the affect ton `this`, I changed the generated `ToString()` to create a new instance of the message, wrap it and then use it with `BuildString()`. This way `this` is not affected.

```
public override string ToString()
{
    var sb = new StringBuilder(100);
    var m = new GroupAndVarLength();
    m.WrapForDecode(_buffer, _offset, _actingBlockLength, _actingVersion);
    m.BuildString(sb);
    return sb.ToString();
}
```
 
The assumption is that `ToString()` is only called for logging and debugging purposes, so the extra object created doesn't affect performance.

Ami

### DIFF
--- a/csharp/sbe-tests/Issue992Tests.cs
+++ b/csharp/sbe-tests/Issue992Tests.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Org.SbeTool.Sbe.Dll;
+using Mktdata;
+
+namespace Org.SbeTool.Sbe.Tests
+{
+  [TestClass]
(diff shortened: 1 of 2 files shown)


amibar:
Hi,

Can someone merge this pull request anytime ?

Ami

amibar:
Thanks

PR #995, opened 2024-05-07
[Rust] Generation: Avoid warnings

YuhanunCitgez:
This PR intends to avoid warnings that get generated because of some clippy flags.

We glob re-export  the statics generated within the codecs, which causes issues due to duplicate naming.

In an ideal world, we would stop exporting the statics (or naming them uniquely), but for now I'd like to just avoid the warnings.

Additionally; `get_bytes` is unused and unusable outside of the generated code itself, rendering it useless for any real-world project.


YuhanunCitgez:
No clue why this failed 😓  Only Rust generation has changed. Is this an already failing test? @vyazelenko

vyazelenko:
@YuhanunCitgez Try merging https://github.com/real-logic/simple-binary-encoding/commit/0053ff3e605efb2af3e79a2b20178a6e1f18255a commit into your branch.

YuhanunCitgez:
@vyazelenko shall we merge?

YuhanunCitgez:
Not sure how to push ahead here. 

Pinging @vyazelenko and @mjpt777 :)

PR #996, opened 2024-05-22
[Rust] Make Rust crate version configurable.

gierlachg:
Not sure if this is something you would be willing to accept.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java
@@ -48,6 +48,7 @@ import static uk.co.real_logic.sbe.ir.Signal.BEGIN_SET;
  */
 public class RustGenerator implements CodeGenerator
 {
+    private static final String CRATE_VERSION = System.getProperty("sbe.rust.crate.version");
     static final String WRITE_BUF_TYPE = "WriteBuf";
     static final String READ_BUF_TYPE = "ReadBuf";
     static final String BUF_LIFETIME = "'a";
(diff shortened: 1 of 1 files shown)


sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java
```diff
@@ -117,7 +117,7 @@ public void generate() throws IOException
 
             indent(writer, 0, "[package]\n");
             indent(writer, 0, "name = \"%s\"\n", namespace);
-            indent(writer, 0, "version = \"0.1.0\"\n");
+            indent(writer, 0, "version = \"%s\"\n", ir.semanticVersion());
```
YuhanunCitgez, line 120: This would be great, I +1 this.

Even better would be if we put this in codegen too, so you could start checking for semver against clients?

ZachBray: I believe [using a semanticVersion is optional](https://github.com/real-logic/simple-binary-encoding/blob/a5e12d1c0d22e477ca302ee58c352984540e882f/sbe-tool/src/main/resources/fpl/sbe.xsd#L54-L60) and `ir.semanticVersion()` might be `null`.

ZachBray:
It is common for SBE schemas to use a FIX version name or `$major.$minor`, e.g., in [Aeron Archive's SBE schema](https://github.com/real-logic/aeron/blob/9fb1e8e5ef5ce75ccd8976a161a1085d3040a951/aeron-archive/src/main/resources/archive/aeron-archive-codecs.xml#L6), as the `messageSchema#semanticVersion`. [The SBE specification doesn't specify any particular format for the semanticVersion attribute value](https://www.fixtrading.org/standards/sbe-online/). Therefore, it would be best not to expect a particular format in the code generator.

From all the file changes, I assume you need to output in `x.y.z` format?

Please can you make the code generator more lenient and reduce the number of changes across existing schema files?

You could consider using a system property to specify the crate version instead of using the `semanticVersion` attribute value. I suspect there are cases where the crate version may change, e.g., when you upgrade the SBE code generator, but the schema does not.

gierlachg:
> From all the file changes, I assume you [need to output in x.y.z format](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field)?

Exactly.

> Please can you make the code generator more lenient and reduce the number of changes across existing schema files?

Done.

> You could consider using a system property to specify the crate version instead of using the semanticVersion attribute value. I suspect there are cases where the crate version may change, e.g., when you upgrade the SBE code generator, but the schema does not.

Makes sense.  What about [this](https://github.com/real-logic/simple-binary-encoding/commit/710f6d442feacee0c49f440ebda06a0006f5e3f1)?

PR #1011, opened 2024-08-22
[C] add test to ensure generated C code properly compiles when versio…

nbradac:
…ns are set

### DIFF
--- a/sbe-tool/src/test/c/CComplianceWithVersionTest.c
+++ b/sbe-tool/src/test/c/CComplianceWithVersionTest.c
@@ -13,13 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "code_generation_test/car.h"
-#include "code_generation_test/messageHeader.h"
-
-#include "group_with_data/testMessage1.h"
-#include "group_with_data/testMessage2.h"
-#include "group_with_data/testMessage3.h"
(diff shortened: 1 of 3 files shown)


sbe-tool/src/test/c/CComplianceWithVersionTest.c
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "code_generation_test_version/car.h"
+#include "code_generation_test_version/messageHeader.h"
+
+int main(void)
+{
```
vyazelenko, line 20: Shouldn't this test access anything from `car.h` or `messageHeader.h`, i.e. what kind of compilation check will it be doing?

nbradac: It's validating that the code in the header files actually compiles.  Without the fix that you merged earlier, this file would fail to compile. We'd see a bunch of this when running cppbuild:

```
/Users/natebradac/src/real-logic/simple-binary-encoding/cppbuild/Release/generated/c/code_generation_test_version/car.h:2705:16: error: expected expression
        return { NULL, 0 };
               ^
```

PR #1027, opened 2024-11-20
[C++] Integrate std::span support for flyweight API

mspdt22:
The impetus was a bug that we ran into when writing a string-literal to a fixed-width char field:

```c++
flyweight.putFixedChar("hello");
```

This is unsafe:
- If the field size is less than 6, we overrun the buffer and corrupt it.
- If the field size is more than 6, we don't zero pad the rest of it.

Instead, we build on support for the std::string_view getters and setters, which do length checking. std::span generalizes this to fixed-width fields of all types. Notably, if the size of the std::span is knowable at compile time, we pay no runtime cost for the length checking, and we should get similar performance to the existing API which takes a raw pointer.

Further, we add a sbetool option to disable accepting arrays by raw pointer, which should prevent memcpy operation without bounds checking. This is off by default to avoid a breaking change.


vyazelenko:
#1038 was merged instead.

PR #1038, opened 2024-12-13
Std span

nbradac:
This is an extension of this PR: (https://github.com/real-logic/simple-binary-encoding/pull/1027) that only allows SBE_USE_SPAN to be set when SBE_ENABLE_SPAN is also set.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -1722,14 +1722,16 @@ public class CppGenerator implements CodeGenerator
             "#if __cplusplus >= 201703L\n" +
             "#  include <string_view>\n" +
             "#  define SBE_NODISCARD [[nodiscard]]\n" +
-            "#  define SBE_USE_STRING_VIEW 1\n" +
+            "#  if !defined(SBE_USE_STRING_VIEW)\n" +
+            "#    define SBE_USE_STRING_VIEW 1\n" +
+            "#  endif\n" +
             "#else\n" +
             "#  define SBE_NODISCARD\n" +
             "#endif\n\n" +
 
             "#if __cplusplus >= 202002L\n" +
-            "#  if defined(SBE_ENABLE_SPAN)\n" +
-            "#    include <span>\n" +
+            "#  include <span>\n" +
+            "#  if !defined(SBE_USE_SPAN)\n" +
             "#    define SBE_USE_SPAN 1\n" +
             "#  endif\n" +
             "#endif\n\n" +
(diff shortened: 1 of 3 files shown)


sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
```diff
@@ -1703,10 +1722,18 @@ private CharSequence generateFileHeader(
             "#if __cplusplus >= 201703L\n" +
             "#  include <string_view>\n" +
             "#  define SBE_NODISCARD [[nodiscard]]\n" +
+            "#  define SBE_USE_STRING_VIEW 1\n" +
```
vyazelenko, line 1725: Defines should be protected with `#ifndef XYZ` or `#if !define(XYZ)` like `__STDC_LIMIT_MACROS` below.

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
```diff
@@ -1703,10 +1722,18 @@ private CharSequence generateFileHeader(
...
             "#else\n" +
             "#  define SBE_NODISCARD\n" +
             "#endif\n\n" +
 
+            "#if __cplusplus >= 202002L\n" +
+            "#  if defined(SBE_ENABLE_SPAN)\n" +
+            "#    include <span>\n" +
+            "#    define SBE_USE_SPAN 1\n" +
+            "#  endif\n" +
+            "#endif\n\n" +
```
vyazelenko, line 1735: Given that the `SBE_USE_STRING_VIEW` is defined if compiler has support for `string_view`. We should do the same for `SBE_USE_SPAN`, i.e. no need to have an additional `SBE_ENABLE_SPAN`.
The definition of `SBE_USE_SPAN` should be protected with `#ifndef`.

sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
```diff
@@ -49,6 +49,8 @@ public class CppGenerator implements CodeGenerator
 {
     private static final boolean DISABLE_IMPLICIT_COPYING = Boolean.parseBoolean(
         System.getProperty("sbe.cpp.disable.implicit.copying", "false"));
+    private static final boolean DISABLE_RAW_ARRAYS = Boolean.parseBoolean(
+        System.getProperty("sbe.cpp.disable.raw.arrays", "false"));
```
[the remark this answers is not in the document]
vyazelenko: @MarcelW-SG This property disables support for raw array encode methods when `std::span` is used.

For example:
* With `sbe.cpp.disable.raw.arrays=false` there is a method which support the raw array:
```c++
    #ifdef SBE_USE_SPAN
    template <typename T>
    ArrayPair &putFalseX(T&& src)  SBE_NOEXCEPT requires
        (std::is_pointer_v<std::remove_reference_t<T>> &&
         !std::is_array_v<std::remove_reference_t<T>>)
    #else
    ArrayPair &putFalseX(const char *const src) SBE_NOEXCEPT
    #endif
    {
        std::memcpy(m_buffer + m_offset + 9, src, sizeof(std::int32_t) * 5);
        return *this;
    }
```
* With `sbe.cpp.disable.raw.arrays=true` this method is not generated.

P.S. I've updated Wiki page.

PR #1061, opened 2025-03-21
[C#] Read NuGet package version from version file

pirrmann:
So that when you generate the NuGet package, even locally, it uses the correct version automatically


vyazelenko:
cc @JPWatson

PR #1062, opened 2025-03-21
[C#] Enable dual build for netstandard2.0 & netstandard2.1

pirrmann:
With this, a consumer of the produced NuGet package does not need to reference the obsolete package System.Memory 4.5.3 anymore, when targeting NETStandard 2.1.
This change still maintains compatibility for consumers who target NETSTandard 2.0, for anyone still running on Full .NET Framework...


vyazelenko:
cc @JPWatson

vyazelenko:
cc @ZachBray

PR #1067, opened 2025-04-10
Return None for optional primite field if its version is less than actingVersion.

marciorasf:
fix issue [!1066](https://github.com/aeron-io/simple-binary-encoding/issues/1066)


vyazelenko:
@marciorasf The PR has merge conflicts after #1058 was merged.

PR #1068, opened 2025-04-26
[Go] Fix compile issue in generated bitset XXChoiceValues struct

zhangw:
Before the fix, the generated bitset-related code looked like this:

type PresenceMapChoiceValue uint8
type PresenceMapChoiceValues struct {
	FooField            PresenceMapChoiceValue
        BarField             PresenceMapChoiceValue
}
var PresenceMapChoice = PresenceMapChoiceValues{uint64(0), uint64(1)}

This caused a compile error:
**cannot use uint64(0) (constant 0 of type uint64) as PresenceMapChoiceValue value in struct literal
compiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)**
The issue is that uint64(0) cannot be directly assigned to a field of type PresenceMapChoiceValue (uint8 alias).

After the fix, the generated code is:
var PresenceMapChoice = PresenceMapChoiceValues{0, 1}

This is simple, correct, and compiles without the issues before.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/golang/struct/GolangGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/golang/struct/GolangGenerator.java
@@ -1650,7 +1650,10 @@ public class GolangGenerator implements CodeGenerator
         String comma = "";
         for (final Token token : tokens)
         {
-            // In Go a composite‐literal’s elements must either be untyped constants (which are assignable to any compatible numeric type) or exactly the same type as the field. So the untyped constants, such 0, 1 works fine, but if using the primite type such as uint64{0} would not works here.
+          // In Go a composite‐literal’s elements must either be untyped constants.
+          //  (which are assignable to any compatible numeric type) or exactly the same type as the field.
+          // So the untyped constants, such 0, 1 works fine,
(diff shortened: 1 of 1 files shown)


vyazelenko:
@zhangw Please add a test for this fix.

vyazelenko:
@zhangw We have [example-schema.xml](https://github.com/aeron-io/simple-binary-encoding/blob/master/sbe-samples/src/main/resources/example-schema.xml) that contains a set definition:
```xml
<set name="OptionalExtras" encodingType="uint8">
    <choice name="sunRoof">0</choice>
    <choice name="sportsPack">1</choice>
    <choice name="cruiseControl">2</choice>
</set>
```
And the code generated for it when you run `./gradlew generateGolangCodecs` (check `gocode/struct/src/extension/OptionalExtras.go`):
```golang
type OptionalExtrasChoiceValue uint8
type OptionalExtrasChoiceValues struct {
	SunRoof       OptionalExtrasChoiceValue
	SportsPack    OptionalExtrasChoiceValue
	CruiseControl OptionalExtrasChoiceValue
}

var OptionalExtrasChoice = OptionalExtrasChoiceValues{0, 1, 2}
```

PR #1092, opened 2025-10-30
[C#] fix: handle case when valueRef is with enum type prefix

unsafePtr:
## Background

Tried to generate models from [Binace Spot spec](https://raw.githubusercontent.com/binance/binance-spot-api-docs/refs/heads/master/sbe/schemas/spot_3_1.xml) which got following line `<field id="1" name="filterType" type="filterType" presence="constant" valueRef="filterType.TrailingDelta"/>`

This is producing incorrect code for c# where whole valueRef is considered as constant. In contrast this case is handled [in C++](https://github.com/aeron-io/simple-binary-encoding/blob/master/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java#L3553).  or in [Rust](https://github.com/aeron-io/simple-binary-encoding/blob/master/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/rust/RustGenerator.java#L980).


vyazelenko:
@unsafePtr Thanks for the fix. It is available in [1.36.1](https://github.com/aeron-io/simple-binary-encoding/releases/tag/1.36.1) release.

PR #1095, opened 2026-02-12
[Rust] Generate setters for optional primitive fields

leonardoarcari:
## Problem

The Rust codegen has no actionable way to set optional scalar primitive fields to their `null`-value, which is only made available as text in the generated docstring. Moreover, there's no uniform way to `null`-initialize all optional primitive scalar fields at the Encoder level. 

In practice, callers have to remember and invoke each optional field setter individually with its null representation, which is error-prone, harder to audit, and easy to miss as schemas evolve.

## Design Proposal

The generated encoders now expose a dedicated nullification path for optional primitive scalar fields:

- A generated method that clears all optional primitive scalar fields in one place (inspired by [sbepp](https://oleksandrkvl.github.io/sbepp/1.7.0/examples.html#autotoc_md67) approach).
- Per-field optional-friendly setters for optional primitive scalar fields, so generated nullification logic can remain schema-driven and centralized.
- For message/composite/group scopes with no optional primitive scalar fields, the generated nullification method is still present as an intentional no-op with explicit documentation in the body (principle of least-surprise)

### Generated API Examples

```rust
// Existing required-value setter remains unchanged.
pub fn cup_holder_count(&mut self, value: u8) { ... }

// New additive optional-friendly setter.
pub fn cup_holder_count_opt(&mut self, value: Option<u8>) {
    match value {
        Some(value) => self.cup_holder_count(value),
        None => self.cup_holder_count(0xff_u8), // schema null value
    }
}
```

```rust
// Generated when optional primitive scalar fields exist.
pub fn nullify_optional_fields(&mut self) {
    self.cup_holder_count_opt(None);
    self.some_other_optional_field_opt(None);
}
```

```rust
// Generated when no optional primitive scalar fields exist in that scope.
pub fn nullify_optional_fields(&mut self) {
    // No optional primitive scalar fields exist; nothing to do.
}
```

This design improves ergonomics and correctness for callers that need deterministic “absent” encoding semantics, while keeping generated API behavior explicit across all encoder scopes.

## Compatibility Decision

I considered changing the existing required-value setter signatures to accept `Option<T>` directly.  
I did not take that path because it would be a breaking API change for existing generated-code consumers and would force updates across downstream codebases.

Instead, I introduced additive methods so existing call sites continue to compile unchanged, and clients can adopt optional-aware behavior incrementally.

## Test Coverage Summary

The test suite covers:

- Optional field setters encoding both present and absent values correctly.
- End-to-end nullification on message encoders with optional primitive fields.
- End-to-end nullification on composite encoders with optional primitive fields.
- End-to-end nullification on group encoders with optional primitive fields.
- No-op behavior on message/composite/group encoders that have no optional primitive fields.
- Regression checks that decoding after nullification returns `None` where expected.

### DIFF
--- a/rust/tests/optional_field_nullify_test.rs
+++ b/rust/tests/optional_field_nullify_test.rs
@@ -123,32 +123,37 @@ fn encode_extension_car_for_optional_field_test(
     );
     car = car.header(0).parent()?;
 
-    car.cup_holder_count_opt(cup_holder_count);
     if nullify {
         // When enabled, tests expect optional message fields to be encoded as "absent".
-        car.nullify_optional_fields();
+        car.cup_holder_count_opt(cup_holder_count)
+            .nullify_optional_fields();
     }
-    car.serial_number(1234);
-    car.model_year(2013);
-    car.available(ExtBooleanType::T);
-    car.code(ExtModel::A);
-    car.some_numbers(&[0, 1, 2, 3]);
-    car.vehicle_code(b"abcdef");
+    else
+    {
+        car.cup_holder_count_opt(cup_holder_count);
+    }
+    car.serial_number(1234)
+        .model_year(2013)
+        .available(ExtBooleanType::T)
+        .code(ExtModel::A)
+        .some_numbers(&[0, 1, 2, 3])
+        .vehicle_code(b"abcdef");
 
-    extras.set_cruise_control(true);
-    extras.set_sports_pack(true);
-    extras.set_sun_roof(false);
+    extras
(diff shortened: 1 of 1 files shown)


vyazelenko:
@mward Can you please review this PR? Thanks.

leonardoarcari:
> I'll take a look

Appreciate it, @mward!

mward:
I'll take a look at changes

leonardoarcari:
Hey @mward! Any feedback on recent changes? Otherwise the PR is complete from my point of view

PR #1099, opened 2026-02-18
[C++] Generate safe getters for optional fields

leonardoarcari:
### Summary
This MR adds safe optional accessors to generated C++ codecs for optional primitive and enum fields.

### Problem
C++ generated decoders expose raw getters that return sentinel null values for optional fields. Callers had to remember field-specific null semantics (including floating-point NaN handling and enum null values), which is error-prone and inconsistent with safer optional-style APIs.

### Design Proposal
The generator now emits additional `*Opt()` accessors for optional fields:
- Optional primitive scalar fields get `std::optional<T> fieldOpt() const`.
- Optional enum fields get `std::optional<Enum::Value> fieldOpt() const`.
- Emission is gated by C++17

```cpp
// Primitive optional field

#if __cplusplus >= 201703L
SBE_NODISCARD std::optional<float> rootFloatOpt() const SBE_NOEXCEPT
{
    const float value = rootFloat();
    const float nullValue = rootFloatNullValue();
    if (std::isnan(value) || value == nullValue)
    {
        return std::nullopt;
    }

    return value;
}
#endif

// Enum optional field

#if __cplusplus >= 201703L
SBE_NODISCARD std::optional<OptionalEnum::Value> rootEnumOpt() const
{
    const OptionalEnum::Value value = rootEnum();
    if (OptionalEnum::NULL_VALUE == value)
    {
        return std::nullopt;
    }

    return value;
}
#endif
```

### Compatibility Decision
The change is additive and non-breaking for generated APIs:
- Existing raw getters/setters are unchanged.
- New optional accessors are emitted only when field presence is optional.
- C++17 guard ensures no impact on older language targets (no `std::optional` API emitted there).

Alternative considered:
- Replace existing raw accessors return type with `std::optional<T>` if building with C++17. Went the safest path not to break compilation in existing callers, but I'm open to suggestions. Replacing return type would make optional field accessors bullet proof in the long run. 

### Test Summary

To validate behavior end-to-end, this PR adds:
- `optional-accessors-schema.xml` containing optional fields in message, group, and composite (including enums).
- `OptionalAccessorsCodeGenTest.cpp` asserting:
  - null sentinels decode to empty optionals,
  - non-null values decode to populated optionals,
  - raw getters remain unchanged.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/generation/cpp/CppGenerator.java
@@ -1733,6 +1733,7 @@ public class CppGenerator implements CodeGenerator
 
             "#if __cplusplus >= 201703L\n" +
             "#  include <string_view>\n" +
+            "#  include <optional>\n" +
             "#  define SBE_NODISCARD [[nodiscard]]\n" +
             "#  if !defined(SBE_USE_STRING_VIEW)\n" +
             "#    define SBE_USE_STRING_VIEW 1\n" +
@@ -1753,6 +1754,7 @@ public class CppGenerator implements CodeGenerator
             "#endif\n\n" +
 
             "#include <cstdint>\n" +
+            "#include <cmath>\n" +
             "#include <limits>\n" +
             "#include <cstring>\n" +
             "#include <iomanip>\n" +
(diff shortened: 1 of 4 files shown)


pluto699:
Also will be good to add method like `fieldNamePresent()` with correct checking for null-value

PR #1101, opened 2026-03-12
[Java] using packed header in header encoders

vachagan-balayan-bullish:
I'm curious why not simply precompute the header for the encoder into a `uint64` and do a single putLong instead of 4 x putShort + with number of bitshifts?

Header is constant for a specific generated encoder... so why compute it every time?


vachagan-balayan-bullish:
regarding matching on `messageHeader` its just an example, obviously you can match according to whatever custom name anyone wants to use... regarding SBE1/2 also i see no issue, fundamentally for all Encoders, once the java class is gneerated its a constant what goes into headers... so can be written in 1 instruction.

PR #1111, opened 2026-06-29
catch validation errors and return -1 instead of propagating an exception

nbradac:
Requested here: https://github.com/aeron-io/simple-binary-encoding/issues/1106.

Straightforward enough to change the behavior.  Not sure if, elsehwere, we're depending on that exception being thrown.

### DIFF
--- a/sbe-tool/src/main/java/uk/co/real_logic/sbe/SbeTool.java
+++ b/sbe-tool/src/main/java/uk/co/real_logic/sbe/SbeTool.java
@@ -285,7 +285,7 @@ public class SbeTool
                 {
                     schema = parseSchema(fileName);
                 }
-                catch (final IllegalStateException | IllegalArgumentException ex)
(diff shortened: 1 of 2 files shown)


sbe-tool/src/main/java/uk/co/real_logic/sbe/SbeTool.java
```diff
@@ -273,7 +280,18 @@ public static void main(final String[] args) throws Exception
                     validateAgainstSchema(fileName, xsdFilename);
                 }
 
-                final MessageSchema schema = parseSchema(fileName);
+                final MessageSchema schema;
+                try
+                {
+                    schema = parseSchema(fileName);
+                }
+                catch (final IllegalStateException | IllegalArgumentException ex)
```
mikeb01: I think could be a more general catch(Exception).

